PR #4, opened 2022-06-17
PI-275 initial project structure and layout for integration tests and sharin…

anthony-britton-moj:
…g data between dev, test and integration test

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,18 +57,18 @@ To build the entire repository using Gradle, run:
 
 To build just a specific project.
 ```shell
-./gradlew <project-name> build
+./gradlew <project-name>:build
 ```
 
 Use buildDependents to build and test all projects that depend on a given project (for instance a shared library)
 ```shell
-./gradlew <project-name> buildDependents
+./gradlew <project-name>:buildDependents
 ```
 
 ## Docker
 To build Docker images and push to your local repository, run:
 ```shell
-./gradlew <project-name> jibDockerBuild
+./gradlew <project-name>:jibDockerBuild
 ```
 
 # Run


marcus-bcl:
Looks good 👍  are you planning on completing the rest of the tests in this branch/pr, or merging changes in incrementally?

Also, might be worth prefixing the commit with PI-275 so it links up with Jira 😄

projects/prison-case-notes-to-probation/src/test/resources/application-test.yml
```diff
@@ -0,0 +1,17 @@
+spring:
+  datasource:
+    url: jdbc:h2:file:./projects/prison-case-notes-to-probation/test;MODE=Oracle;AUTO_SERVER=true;AUTO_SERVER_PORT=9092
```
marcus-bcl, line 3: Should this file be in `src/integrationTest` rather than `src/test`?  I'm guessing we wouldn't need H2 or Hoverfly for the unit tests?

settings.gradle.kts
```diff
@@ -9,6 +9,8 @@ dependencyResolutionManagement {
         create("libs") {
             library("hawtio", "io.hawt:hawtio-springboot:2.15.0")
             library("hoverfly", "io.specto:hoverfly-java:0.14.2")
+            library("hoverfly-junit", "io.specto:hoverfly-java-junit5:0.14.0")
```
marcus-bcl, line 12: Any reason not to use the latest (0.14.2) here?

README.md
```diff
@@ -51,13 +51,13 @@ To set up your development environment,
 # Build
 IntelliJ will automatically build your code as needed.  To build and unit test using Gradle, run:
 ```shell
-./gradlew build
+./gradlew <project-name> build
```
marcus-bcl, line 54: Do we need to specify the project name here?  Leaving it out feels like a simpler starting point for someone first reading about the project - maybe have a separate line/section showing how to build everything vs building a specific sub-project?

Also, should it be `./gradlew :<project-name>:build`?

projects/prison-case-notes-to-probation/src/test/resources/application-test.yml
marcus-bcl: Maybe better to keep it consistent so `application-integration-test.yml` - happy either way though.

README.md
marcus-bcl: But it still needs the colon between the project name and the task name doesn't it?

marcus-bcl (approved):

PR #6, opened 2022-06-21
updated db ur for dev and test file based dbs

anthony-britton-moj:


marcus-bcl (approved):
Will this affect running from root via Gradle?

PR #30, opened 2022-06-28
separated logic for jib config into plugin

anthony-britton-moj:


marcus-bcl (approved):
Nice, I like it!  So we just conditionally enable the jib plugin in each project and it'll be pre-configured by the buildSrc class?

PR #32, opened 2022-06-28
Ab classpath jacoco config

anthony-britton-moj:


marcus-bcl (approved):
Looks good.  

It doesn't need to be resolved in this PR, but what do you think about splitting out the commons library at some point?  It feels a bit generic right now, and we might be forced to pull in things we don't want/need.  Could it be broken up into (for example) a separate date library, audit library, telemetry library, etc?

PR #89, opened 2022-08-05
PI-367

anthony-britton-moj:
Turned off hibernate SQL Exception Handler loggers to prevent filling the log with expected constraint violations - they are being caught and dealt with - so no need to log them


marcus-bcl (approved):
Might be worth adding a comment or a commit message to say why we did this.  We'll never remember when we look back in a few months time 😄

PR #95, opened 2022-08-09
PI-291: ADR for building container images using Jib

mxco86:


marcus-bcl (dismissed):
Nice!

Maybe worth mentioning that Jib doesn't need a Docker daemon process to be running - it builds and pushes directly from the Gradle JVM process, which is easier and faster on CI.

Also we're no longer running our Gradle build in a Docker stage, so caching etc works as expected (which used to be a problem for Community API etc).

mxco86:
> Maybe worth mentioning that Jib doesn't need a Docker daemon process to be running - it builds and pushes directly from the Gradle JVM process, which is easier and faster on CI.
> 
> Also we're no longer running our Gradle build in a Docker stage, so caching etc works as expected (which used to be a problem for Community API etc).

Thanks - have added these

marcus-bcl (approved):

PR #117, opened 2022-08-17
PI-370

anthony-britton-moj:


stevomcallister (approved):

.github/workflows/check-roles.yml
```diff
@@ -0,0 +1,61 @@
...
+          echo "::set-output name=client_id_name::${project_name}_CLIENT_ID"
+          echo "::set-output name=client_secret_name::${project_name}_CLIENT_SECRET"
+        working-directory: projects/${{ matrix.project }}/deploy
+
+      - name: Get Hmmps Auth Roles
+        id: hmpps-auth-request
+        env:
+          CLIENT_ID_NAME: ${{ steps.hmpps-auth-values.outputs.client_id_name }}
+          CLIENT_SECRET_NAME: ${{ steps.hmpps-auth-values.outputs.client_secret_name }}
+        uses: fjogeleit/http-request-action@v1
```
marcus-bcl, line 42: We should probably use a specific sha for this - otherwise we're potentially exposing our secrets to `fjogeleit` if they retag their repo with some malicious code.

PR #120, opened 2022-08-18
Upgraded Spring Boot to 2.7.3

anthony-britton-moj:


marcus-bcl:
🤔 wonder why this wasn't picked up by dependabot?

marcus-bcl (approved):

PR #144, opened 2022-09-05
changed probation search url

stevomcallister:


stevomcallister:
> Will this URL conflict with the actual search API, just thinking if we end up renaming that from `probation-offender-search` to `probation-search` at some point?

yes we've thought about that but figured we'd just rename them once it's all working.

PR #154, opened 2022-09-08
Re-enabled Logstash healthcheck

marcus-bcl:


stevomcallister (changes requested):
don't think this works

projects/person-search-index-from-delius/deploy/template/task-definition.yml
```diff
@@ -5,7 +5,7 @@ containerDefinitions:
   - name: {{ .values.image.name }}
     image: ghcr.io/ministryofjustice/hmpps-probation-integration-services/{{ .values.image.name }}:{{ .Env.VERSION }}
     healthCheck:
-      command: ["CMD-SHELL", "exit 0"]
+      command: ["CMD-SHELL", "curl -sf localhost:9600"]
```
stevomcallister, line 8:
```suggestion
      command: ["CMD-SHELL", "curl -sSf localhost:9600"]
```

stevomcallister (changes requested):

marcus-bcl:
> don't think this works

I deployed the change via the pipeline here: https://github.com/ministryofjustice/hmpps-probation-integration-services/actions/runs/3013509248

The containers started up and reported healthy, so it should be okay to merge.

projects/person-search-index-from-delius/deploy/template/task-definition.yml
marcus-bcl: Even if we show the error with `-S`, it'll get swallowed by ECS anyway won't it?  Happy to make the change though if you think it's required?

marcus-bcl: For my own reference as I had to look it up 😄
 https://explainshell.com/explain?cmd=curl+-sSf+localhost%3A9600

stevomcallister (approved):

PR #176, opened 2022-09-15
added more trivy ignores

anthony-britton-moj:


marcus-bcl:
We'll only get alerted for HIGH/CRITICAL, do we need suppressions for the MEDIUM severity vulnerabilities?

PR #200, opened 2022-09-29
PI-317 Add project templates to bootstrap

marcus-bcl:
Test run here: https://github.com/ministryofjustice/hmpps-probation-integration-services/actions/runs/3149658524

The bootstrap created this pull request for a "message-listener": https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/201

### DIFF
--- a/.github/workflows/bootstrap.yml
+++ b/.github/workflows/bootstrap.yml
@@ -8,6 +8,15 @@ on:
         description: Project name
         required: true
         type: string
+      project_template:
+        description: Project template
+        required: true
+        type: choice
+        default: message-listener-with-api-client
+        options:
+          - message-listener-with-api-client
+          - message-listener
+          - 'No template - I want to create the project from scratch'
       create_database_credentials:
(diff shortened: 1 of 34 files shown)


marcus-bcl:
> Initially I thought it was going to be message listener or message listener with incoming api requests to distinguish the two types of service we have? Here it appears that the message-listener with api client is the former. However, not sure how often (if ever) we will have a service without the api client since the events are always 'lightweight' and usually have to call the detail url? Will there be a third type added for incoming API requests and message handling -- eg Pre-Sentence Reports?

Yes we should add a template for message-listener-with-api-server (or something, I couldn't think of good names for these 😅).

Technically yes our message listeners should always be making an API call to a detailUrl, however that's currently not the case for `prison-custody-status-to-delius`, and for this new one `risk-assessment-scores-to-delius`.  So I decided to create a simple template without the api-client, that I could test when creating the `risk-assessment-scores-to-delius` project.

@anthony-britton-moj could I get another review?

PR #239, opened 2022-10-11
Add multiplatform OpenJDK image

pezholio:
After trying in https://github.com/ministryofjustice/community-api/pull/732 to switch to a multiplatform build, it became apparent that there are no arm64 17 jre alpine Docker images, and also the underlying image in this repo is not multiplatform. This adds a new multiplatform image based on the version 18 Eclipse Temurin image, which is multiplatform, and used in production via the Kotlin template.

### DIFF
--- a/tools/eclipse-temurin/19-jre-jammy/Dockerfile
+++ b/tools/eclipse-temurin/19-jre-jammy/Dockerfile
@@ -1,0 +1,2 @@
+FROM --platform=$BUILDPLATFORM eclipse-temurin:19-jre-jammy
+RUN apt-get update && apt-get upgrade -y && apt-get clean && rm -rf /var/lib/apt/lists/*
--- a/tools/eclipse-temurin/19-jre-jammy/build.gradle.kts
+++ b/tools/eclipse-temurin/19-jre-jammy/build.gradle.kts
@@ -1,0 +1,10 @@
+apply(plugin = "com.palantir.docker")
+
+tasks {
+    configure<com.palantir.gradle.docker.DockerExtension> {
+        name = "ghcr.io/ministryofjustice/hmpps-probation-integration-services/eclipse-temurin:19-jre-jammy"
+        tag("19-jre-jammy", "ghcr.io/ministryofjustice/hmpps-probation-integration-services/eclipse-temurin:19-jre-jammy")
+        buildx(true)
+        platform("linux/arm64", "linux/amd64")
+    }
+}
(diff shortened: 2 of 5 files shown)


pezholio:
@anthony-britton-moj I have now!

tools/settings.gradle.kts
```diff
@@ -1,4 +1,5 @@
 include(
     "eclipse-temurin:17-jre",
-    "opensearch-logstash:8",
+    "eclipse-temurin:18-jre-jammy",
```
marcus-bcl, line 3: I think this needs to be updated to 19 as well, to be picked up by the build

marcus-bcl (changes requested):

tools/eclipse-temurin/19-jre-jammy/build.gradle.kts
```diff
@@ -0,0 +1,10 @@
+apply(plugin = "com.palantir.docker")
+
+tasks {
+    configure<com.palantir.gradle.docker.DockerExtension> {
+        name = "ghcr.io/ministryofjustice/hmpps-probation-integration-services/eclipse-temurin:19-jre-jammy"
+        tag("19-jre-jammy", "ghcr.io/ministryofjustice/hmpps-probation-integration-services/eclipse-temurin:19-jre-jammy")
+        buildx = true
+        platform = arrayOf("linux/arm64", "linux/amd64")
```
marcus-bcl, line 8:
```suggestion
        buildx(true)
        platform("linux/arm64", "linux/amd64")
```
The Gradle Kotlin DSL isn't very intuitive, but this should work 😄

tools/eclipse-temurin/19-jre-jammy/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM --platform=$BUILDPLATFORM eclipse-temurin:19-jre-jammy
+RUN apk upgrade --no-cache && rm -rf /var/cache/apk/*
```
marcus-bcl, line 2: `apk` is for Alpine, for Ubuntu you need `apt` or `apt-get`:
```suggestion
RUN apt-get update && apt-get upgrade -y && apt-get clean && rm -rf /var/lib/apt/lists/*
```

pezholio: Ah, good catch. That's a side effect of me trying an Alpine image before

marcus-bcl (approved):

PR #241, opened 2022-10-11
Fix pushing multiplatform images

marcus-bcl:


pezholio:
I think https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/242 may be what we need

pezholio (approved):
This looks good. I don't think we need those first two commits now - they can probably be rebased away

PR #246, opened 2022-10-12
PI-477 check if value changed before updating aws

anthony-britton-moj:


.github/workflows/deploy.yml
```diff
@@ -94,8 +94,11 @@ jobs:
           echo "${ALL_SECRETS}" | jq -r ". | keys[] | select(. | startswith(\"${project_name}\"))" | while read -r key; do
             name=$(echo "${key}" | tr '[:upper:]' '[:lower:]' | sed 's/_/-/g' | sed 's/${{ matrix.project }}-//')
             value=$(echo "${ALL_SECRETS}" | jq -r ".${key}")
-            echo "Adding secret to parameter store: ${key} (${namespace}/${name})"
-            aws ssm put-parameter --name "${namespace}/${name}" --value "${value}" --overwrite --type SecureString --region eu-west-2
+            existing=$(aws ssm get-parameter --name "${namespace}/${name}")
```
marcus-bcl, line 97: Think you might need to add the `--region eu-west-2`.  Also maybe `--with-decryption`.

PR #264, opened 2022-10-17
PI-488 Create ADR for managing secrets using GitHub Environments

mxco86:


marcus-bcl (approved):
Is it worth mentioning that we push the secrets to AWS parameter store?  I guess that will change when we move to CP anyway.

mxco86:
> Is it worth mentioning that we push the secrets to AWS parameter store? I guess that will change when we move to CP anyway.

Yes I left this out as it's tied to the deployment platform which may be different for different services

PR #281, opened 2022-10-20
PI-500

anthony-britton-moj:
Extracted messaging libs into separate lib from commons
Updated templates to include new lib


marcus-bcl:
Do we need to break up the TelemetryService, so the `trackEvent` method can still be used by non-messaging services?  We could have the simple TelemetryService in commons, and define some extension functions in messaging?

marcus-bcl:
@anthony-britton-moj nice!

marcus-bcl (approved):

PR #301, opened 2022-10-25
PI-503

anthony-britton-moj:


projects/approved-premises-and-oasys/deploy/.helmignore
```diff
@@ -0,0 +1,21 @@
+# Patterns to ignore when building packages.
```
marcus-bcl, line 1: Do we need this file any more?

.github/workflows/deploy-old.yml
```diff
@@ -0,0 +1,159 @@
+name: Deploy
```
marcus-bcl, line 1: Assuming we can delete this file now?

projects/approved-premises-and-oasys/deploy/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+appVersion: '1.0'
+description: A Helm chart for Kubernetes
+name: approved-premises-and-oasys
+version: 1.0.0
+
+dependencies:
+  - name: generic-service
+    version: 1.6.0
+    repository: https://ministryofjustice.github.io/hmpps-helm-charts
```
marcus-bcl, line 10: Shall we try adding the "generic-prometheus-alerts" in now too?

projects/approved-premises-and-oasys/deploy/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v2
+appVersion: '1.0'
```
marcus-bcl, line 2: According to https://helm.sh/docs/topics/charts/#the-appversion-field, this field corresponds to the version of the app, so I think doing the `sed`/`yq` to set it would be a good idea.  What do you think? 

Also the other `version` field is the chart version, and should follow semver, so maybe we just leave it as 1.0.0 given we won't really be changing this file - i.e. remove the `--version` flag in the deployment ?

projects/approved-premises-and-oasys/deploy/values.yaml
```diff
@@ -0,0 +1,28 @@
...
+    tlsSecretName: approved-premises-and-oasys-cert
+
+  env:
+    APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.json
+
+  namespace_secrets:
+    approved-premises-and-oasys:
+      APPLICATIONINSIGHTS_CONNECTION_STRING: APPLICATIONINSIGHTS_CONNECTION_STRING
+      SENTRY_DSN: SENTRY_DSN
+      ORDS_OAUTH2_CLIENT-ID: ORDS_OAUTH2_CLIENT
```
marcus-bcl, line 23:
```suggestion
      ORDS_OAUTH2_CLIENT-ID: ORDS_OAUTH2_CLIENT-ID
```
is this a typo?

Looks good otherwise.

marcus-bcl (approved):

PR #303, opened 2022-10-26
PI-521 changed default date format

stevomcallister:


marcus-bcl (commented):
There are a few instances of hard-coding the "dd/MM/yyyy HH:mm:ss" format - could we replace those with the DeliusDateTimeFormatter now?

E.g. https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/820a33d6aefbf61c42f88bb088d1c9b1cb612191/projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/manager/responsibleofficer/ResponsibleOfficer.kt#L73

marcus-bcl (approved):

PR #324, opened 2022-10-31
PI-247 Schema spy report GH action

stevomcallister:


.github/workflows/schema-spy.yml
```diff
@@ -0,0 +1,38 @@
+name: Schema Spy report generation
+
+on:
+  schedule:
+    - cron: "0 7 * * TUE" # Every tuesday at 07:00 UTC
```
marcus-bcl, line 5: Does this run against the delius-test environment?  If so, we should probably change this to later as the deployment Delius deployment usually happens around midday on a Tuesday.

Also is it worth running this whenever we push changes to the schema spy config?  To make sure it still works if we change config or version etc:
```suggestion
  schedule:
    - cron: "0 7 * * TUE" # Every tuesday at 07:00 UTC
  push:
    paths:
      - schema-spy/*
      - .github/workflows/schema-spy.yml
```

.github/workflows/schema-spy.yml
```diff
@@ -0,0 +1,38 @@
...
+          PORT: ${{ secrets.SCHEMA_SPY_PORT }}
+          SCHEMA: ${{ secrets.SCHEMA_SPY_SCHEMA }}
+          USERNAME: ${{ secrets.SCHEMA_SPY_USERNAME }}
+          PASSWORD: ${{ secrets.SCHEMA_SPY_PASSWORD }}
+        run: run.sh
+        working-directory: schema-spy
+
+      - name: Publish HTML report
+        uses: JamesIves/github-pages-deploy-action@v4
+        if: always()
```
marcus-bcl, line 30: Don't think we want this line?  If the SchemaSpy generation fails, should we still try to publish the report?

.github/workflows/schema-spy.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          folder: schema-spy/DeliusSchema
+          target-folder: schema-spy/${{ github.ref_name }}/${{ github.sha }}
+
+      - name: Add HTML report URL to the job summary
+        if: always()
```
marcus-bcl, line 36: Same here.

marcus-bcl:
Was there a problem putting this in the tools directory btw?  Feels like that was a better place for it?

PR #326, opened 2022-11-01
PI-511 Added new CVEs for xstream suppression

izzyL-DJ:


projects/person-search-index-from-delius/.trivyignore
```diff
@@ -21,4 +21,9 @@ CVE-2022-42004
 
 # Suppression for Xstream to serialise XML data was vulnerable to Denial of Service attacks
 # can be suppressed as  it’s not receiving data from untrusted source
```
marcus-bcl, line 23: I don't think this is needed here is it?  The scan for person-search-index-to-delius is passing, as it doesn't have this dependency.

marcus-bcl (approved):

PR #341, opened 2022-11-04
PI-540 add domain events endpoint for breach details

markheyward-chaplin:


projects/domain-events-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/Nsi.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    @ManyToOne
+    @JoinColumn(name = "event_id", updatable = false)
+    val event: Event? = null,
+
+    @Column(name = "active_flag", updatable = false, columnDefinition = "NUMBER")
+    val active: Boolean = true,
+
+    @Column(updatable = false, columnDefinition = "NUMBER")
+    var softDeleted: Boolean = false,
```
marcus-bcl, line 37: Any reason for this to be a `var`? Also it doesn't appear to be used anywhere?

projects/domain-events-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/Nsi.kt
```diff
@@ -0,0 +1,65 @@
...
+    @ManyToOne
+    @JoinColumn(name = "nsi_status_id", updatable = false)
+    val status: StandardReferenceList,
+
+    @ManyToOne
+    @JoinColumn(name = "event_id", updatable = false)
+    val event: Event? = null,
+
+    @Column(name = "active_flag", updatable = false, columnDefinition = "NUMBER")
+    val active: Boolean = true,
```
marcus-bcl, line 34: Is this being used anywhere?  Can it be removed?

projects/domain-events-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/Nsi.kt
```diff
@@ -0,0 +1,65 @@
...
+import javax.persistence.Table
+
+@Immutable
+@Entity
+class Nsi(
+    @Id
+    @Column(name = "nsi_id")
+    var id: Long,
+
+    val referralDate: LocalDate,
```
marcus-bcl, line 19: We try to use ZonedDateTime where possible, is there a reason for using LocalDate here?  Same in the `BreachDetails` response.

marcus-bcl (approved):

PR #376, opened 2022-11-11
PI-522 PI-529 PI-530 PI-531 PI-532 PI-534 PI-535 PI-537

anthony-britton-moj:


projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/recall/RecallService.kt
```diff
@@ -132,7 +163,7 @@ class RecallService(
                 notes = "Reason for Recall: ${recallReason.description}",
                 staffId = orderManager.staffId,
                 teamId = orderManager.teamId,
-                createdDatetime = recall.createdDatetime,
+                createdDatetime = recall?.createdDatetime ?: recallDateTime,
```
marcus-bcl, line 166: Should this default to `ZonedDateTime.now()` when we're not linking it to a recall?  The contact date is still set to recallDateTime.

marcus-bcl (commented):
Looks good, as discussed though let's not merge this yet.  But we can deploy it from a branch when we want to end-to-end test it.

stevomcallister (approved):

PR #401, opened 2022-11-15
PI-524 Sign commits with the probation-integration-bot GPG key

marcus-bcl:
Test PR showing commit verified: https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/398


mxco86 (approved):
👍

PR #423, opened 2022-11-24
WFP-1465

carlov20:


carlov20:
yeah, sorry. When we were integrating we realised we always needed the current community person manager unless it was unallocated

PR #450, opened 2022-11-29
PI-616

anthony-britton-moj:


marcus-bcl (approved):
So "unknown" corresponds to something like "no disposals, or multiple active unallocated disposals" ?

PR #516, opened 2022-12-07
PI-571 Handle approved premises application submitted events

marcus-bcl:


marcus-bcl:
> Looks fine, but seems to be a drop in coverage

The coverage looks fine (94.7% on new code), it seems Sonar is complaining about duplications - because we're dealing with Contacts again.   The overall duplication will still only be 1.8% though.

PR #571, opened 2022-12-15
PI-648 Improve error logging in Risk Assessment Scores to Delius consumer

marcus-bcl:
* Adds parsing of validation errors from the stored procedure, to better categorise them in Sentry and App Insights
* Adds new method `addToContext` to the TelemetryService, to ensure customDimensions are added to requests and exceptions as well as custom events.
* Adds tests for the RiskScoreService


marcus-bcl:
> Notice end to end tests are now ran against the branch - nice, but does it matter they failed? Not for the same project as the PR -- but since common telemetry was changed I guess it ran them all.

@anthony-britton-moj sorry no I'm just doing some testing from by triggering the pipeline from the branch, to check what appears in Sentry/App Insights.

PR #600, opened 2022-12-20
PI-634 Handle recall recommendation started messages

anthony-britton-moj:


projects/make-recall-decisions-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -10,10 +11,22 @@ import uk.gov.justice.digital.hmpps.telemetry.notificationReceived
 @Component
 class Handler(
     override val converter: NotificationConverter<HmppsDomainEvent>,
+    private val recommendationStarted: RecommendationStarted,
     private val telemetryService: TelemetryService
 ) : NotificationHandler<HmppsDomainEvent> {
     override fun handle(notification: Notification<HmppsDomainEvent>) {
         telemetryService.notificationReceived(notification)
-        TODO("Not yet implemented")
+        notification.message.personReference.findCrn()?.let {
```
marcus-bcl, line 19: It's unlikely to ever happen, but should we error if the message doesn't have a CRN?

projects/make-recall-decisions-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/recommendation/RecommendationStarted.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val contactRepository: ContactRepository
+) {
+
+    fun recommended(crn: String, recommendationUrl: String) {
+        val person = personRepository.getPerson(crn)
+        contactRepository.save(person.recommendationStarted(recommendationUrl))
+    }
+
+    private fun Person.recommendationStarted(recommendationUrl: String): Contact {
+        val now = ZonedDateTime.now()
```
marcus-bcl, line 27: Should this be the ocurredAt field in the message, rather than the system time?

libs/dev-tools/src/main/kotlin/uk/gov/justice/digital/hmpps/ReflectionUtils.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps
+
+fun Any.set(field: String, value: Any) {
```
marcus-bcl, line 3: Fyi there is a Spring utils method for this: 
`org.springframework.data.util.ReflectionUtils.setField(f, this, value)`

This use case simple enough so happy to leave it for now, but if we start using reflection more in future it might be worth leaning on the Spring classes, as they appear to handle some edge cases and throw better exceptions.

projects/make-recall-decisions-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/recommendation/contact/entity/Contact.kt
```diff
@@ -0,0 +1,123 @@
...
+    val alert: Boolean? = false,
+
+    @Column(updatable = false, columnDefinition = "NUMBER")
+    val trustProviderFlag: Boolean = false,
+
+    @Column(updatable = false)
+    val trustProviderTeamId: Long = teamId,
+
+    @Column(updatable = false)
+    val staffEmployeeId: Long = staffId,
```
marcus-bcl, line 74: I think you can omit this and the trustProvider fields, as they're set by a trigger in the database.

projects/make-recall-decisions-and-delius/deploy/database/access.yml
```diff
@@ -5,5 +5,5 @@ database:
```
marcus-bcl, line 5: Think you'll need to add the table access here to enable writing to the contact / contact_alert tables

marcus-bcl (approved):

PR #605, opened 2022-12-20
PI-636 Move DB user creation to a separate workflow

marcus-bcl:
* The new workflow runs whenever the access.yml file changes, instead of on deployment
* Audit user creation is disabled in preprod, this will be populated by a DB trigger when prod is updated

Related PR: https://github.com/ministryofjustice/hmpps-delius-pipelines/pull/558

Test run: https://github.com/ministryofjustice/hmpps-probation-integration-services/actions/runs/3741802757


bill-buchan (approved):
LGTM

PR #638, opened 2023-01-03
PI-681 replace artemis

anthony-britton-moj:


marcus-bcl (commented):
Looks good, happy to approve but I think you'll need to update the integration tests in the templates directory too.

marcus-bcl (approved):

PR #670, opened 2023-01-06
PI-652 new staff details endpoint

stevomcallister:


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/StaffService.kt
```diff
@@ -0,0 +1,26 @@
...
+@Service
+class StaffService(
+    private val staffRepository: StaffRepository,
+) {
+    fun getOfficerView(code: String): OfficerView {
+        val staff = staffRepository.findByCode(code) ?: throw NotFoundException("Staff", "code", code)
+        return OfficerView(
+            code,
+            Name(staff.forename, staff.middleName, staff.surname),
+            staff.grade?.description,
```
marcus-bcl, line 19: There are a couple of extension functions for these:
```suggestion
            staff.name(),
            staff.grade(),
```

projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/StaffService.kt
```diff
@@ -0,0 +1,26 @@
...
+            null,
+            staffRepository.getSentencesDueCountByStaffId(staff.id, LocalDate.now().minusWeeks(4)),
```
marcus-bcl, line 21: Should this be `dateTo = LocalDate.now().plusWeeks(4)`?

projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/provider/StaffRepository.kt
```diff
@@ -18,4 +19,72 @@ interface StaffRepository : JpaRepository<Staff, Long> {
...
+       select 
+        count(1) 
+        from offender_manager om 
+                join event e on e.offender_id = om.offender_id
+                join disposal d on d.event_id = e.event_id
+                join custody c on d.disposal_id = c.disposal_id
+                join Institutional_report i on i.custody_id = c.custody_id
+                join r_standard_reference_list r on i.institution_report_type_id = r.standard_reference_list_id
+                where om.allocation_staff_id = :staffId
+                and om.end_date is null
```
marcus-bcl, line 77: Would it be better to check `om.active_flag = 1` here?

projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/OfficerView.kt
```diff
@@ -0,0 +1,16 @@
...
+    val code: String,
+    val name: Name,
+    val grade: String?,
+    val email: String?,
+    val casesDueToEndInNext4Weeks: Long,
+    val releasesWithinNext4Weeks: Long,
+    val paroleReportsToCompleteInNext4Weeks: Long,
+)
+
+data class Team(
```
marcus-bcl, line 13: Is this needed?

marcus-bcl (changes requested):
Looks good, think the logic is the wrong way round for "things due in the next four weeks" though.

Also probably worth adding an integration test?  The native queries should work with H2 too.

projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/OfficerView.kt
stevomcallister: nope

projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/provider/StaffRepository.kt
stevomcallister: ok

marcus-bcl (approved):
Looks good, and makes sense that the type is CUSTODY now, since a custody record is being added by the data loader.

PR #673, opened 2023-01-09
PI-647 Probation Record API

anthony-britton-moj:


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/event/sentence/DisposalRepository.kt
```diff
@@ -0,0 +1,20 @@
...
+    @Query(
+        """
+        select new uk.gov.justice.digital.hmpps.integrations.delius.event.sentence.SentenceWithManager(
+            d, mo, s
+        ) from Disposal d
+        join MainOffence mo on d.event.id = mo.event.id
+        join OrderManager om on d.event.id = om.eventId
+        join Staff s on om.staff.id = s.id
+        where d.event.person.id = :personId
+        and d.event.number <> :eventNumber
```
marcus-bcl, line 16: Minor nitpick - it took me a while to realise why you had this clause here.  Might be worth naming the method something like `findAllSentencesExcludingEventNumber` to make it clearer?

marcus-bcl (approved):

marcus-bcl (approved):

PR #678, opened 2023-01-10
PI-697 handle no event number - send detail to telemetry

anthony-britton-moj:


marcus-bcl (changes requested):
The problem here is we're getting invalid JSON and need to ignore it - i.e. catch and drop the JsonMappingException so it stops filling our DLQ with noise.

This is the exception we currently get: https://sentry.io/organizations/ministryofjustice/issues/3799799984

When they eventually switch to passing a null event number then we'll need to handle that and still call the procedure.

libs/messaging/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/NotificationHandler.kt
```diff
@@ -1,12 +1,18 @@
...
+import uk.gov.justice.digital.hmpps.telemetry.TelemetryService
 
 interface NotificationHandler<T : Any> {
     val converter: NotificationConverter<T>
     fun handle(notification: Notification<T>)
     fun handle(message: String) {
-        handle(converter.fromMessage(message))
+        try {
+            handle(converter.fromMessage(message))
+        } catch (e: JsonMappingException) {
```
marcus-bcl, line 14: Does this apply to all services?  Probably makes sense to do this only for risk-assessment-scores-to-delius if possible?

projects/risk-assessment-scores-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -24,7 +24,7 @@ class Handler(
                 "risk-assessment.scores.rsr.determined" -> {
                     riskScoreService.updateRsrScores(
                         message.personReference.findCrn() ?: throw IllegalArgumentException("Missing CRN in ${message.personReference}"),
-                        message.additionalInformation["EventNumber"] as Int,
+                        message.additionalInformation["EventNumber"] as Int? ?: throw DeliusValidationError("No event number provided"),
```
marcus-bcl, line 27: I think we'll want to pass this through if it's null, the stored procedure will then infer which event to use.

marcus-bcl (dismissed):

PR #686, opened 2023-01-11
PI-575 Initial code for handle "person-arrived" messages

marcus-bcl:
To do:
* Answer the questions on the Jira ticket
* Update the data loader to create the NSI type/status/etc for the integration test
* Add code to update the main address


projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/team/TeamRepository.kt
```diff
@@ -5,8 +5,13 @@ import uk.gov.justice.digital.hmpps.exception.NotFoundException
 
 interface TeamRepository : JpaRepository<Team, Long> {
     fun findByCodeAndProbationAreaCode(teamCode: String, probationAreaCode: String): Team?
+    fun findByApprovedPremisesCodeCode(approvedPremisesCode: String): Team?
```
stevomcallister, line 8: codecode?

stevomcallister (approved):

PR #696, opened 2023-01-17
PI-576

anthony-britton-moj:


projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ApprovedPremisesService.kt
```diff
@@ -163,14 +166,38 @@ class ApprovedPremisesService(
...
+            staff = staff,
+            probationAreaCode = details.premises.probationArea.code,
+            createAlert = false
+        )
+        closeNsi(details)
+        endMainAddress(person, details.departedAt.toLocalDate())
+    }
+
+    private fun closeNsi(details: PersonDeparted) {
+        val nsi = nsiRepository.findByExternalReference(EXT_REF_BOOKING_PREFIX + details.bookingId)
```
stevomcallister, line 188: Should this also make the NSI inactive? Or is this done by the DB? if so can the active flag in the NSI entity remain a val?

stevomcallister (approved):

stevomcallister (approved):

PR #706, opened 2023-01-19
Update application.yml config for feign clients

anthony-britton-moj:


projects/workforce-allocations-to-delius/src/main/resources/application.yml
```diff
@@ -5,6 +5,7 @@ spring:
   jpa:
     database-platform: org.hibernate.dialect.Oracle12cDialect
     hibernate.ddl-auto: validate
+    show-sql: true
```
marcus-bcl, line 8: Was this meant to be included?

marcus-bcl (commented):
Do we need to update the templates too? - yeah updated -- always forget them

marcus-bcl (approved):

PR #716, opened 2023-01-19
PI-720 refactor staff / staff with user

anthony-britton-moj:


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/provider/Staff.kt
```diff
@@ -51,3 +48,33 @@ class Staff(
     @Transient
     val displayName = listOfNotNull(forename, middleName, surname).joinToString(" ")
 }
+
+@Immutable
+@Entity
+@Table(name = "staff")
+open class Staff(
```
marcus-bcl, line 55: Does this need to be open?

marcus-bcl (approved):
Shame about the Sonar coverage failure, but I think that's only due to the nature of the changes - overall coverage is still 92%.

marcus-bcl (approved):

PR #728, opened 2023-01-23
PI-732 move offender events to CP

anthony-britton-moj:


projects/offender-events-and-delius/deploy/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+  nameOverride: offender-events-and-delius
+
+  image:
+    repository: ghcr.io/ministryofjustice/hmpps-probation-integration-services/offender-events-and-delius
+
+  ingress:
+    tlsSecretName: offender-events-and-delius-cert
+
+  env:
+    DELIUS_DB_USERNAME: OffenderEventsAndDelius
```
marcus-bcl, line 11: Don't think you need this, as it's defined here: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/ab_PI-732_offender_events/projects/offender-events-and-delius/src/main/resources/application.yml#L9

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #750, opened 2023-01-24
PI-730 Unpaid Work Personal Details API

stevomcallister:


build.gradle.kts
```diff
@@ -70,6 +71,7 @@ subprojects {
         plugin(JibConfigPlugin::class.java)
         plugin(ClassPathPlugin::class.java)
         plugin("org.sonarqube")
+        plugin("org.jetbrains.kotlin.kapt")
```
marcus-bcl, line 74: Do we want to apply this to all projects, or just the ones that need it?

Also for info, apparently kapt is in maintenance mode: https://kotlinlang.org/docs/kapt.html.  Hopefully MapStruct can switch over to KSP, but they've had an issue open for a while: https://github.com/mapstruct/mapstruct/issues/2522

projects/unpaid-work-and-delius/build.gradle.kts
```diff
@@ -20,6 +20,8 @@ dependencies {
     implementation(libs.openfeign)
     implementation(libs.sentry)
     implementation(libs.springdoc)
+    implementation("org.mapstruct:mapstruct:1.5.3.Final")
```
marcus-bcl, line 23: Should we move the version up to settings.gradle, in case we want to use it other projects?

projects/unpaid-work-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/personaldetails/entity/PersonalCircumstanceEntity.kt
```diff
@@ -0,0 +1,78 @@
...
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import org.hibernate.annotations.Immutable
+import org.hibernate.annotations.Where
+import org.hibernate.type.YesNoConverter
+import java.time.LocalDate
+
+@Entity
+@Immutable
+@Where(clause = "soft_deleted = 0 and end_date is null")
```
marcus-bcl, line 17: Do we need to account for the end date being in the future?  Same for personal_contact.
```suggestion
@Where(clause = "soft_deleted = 0 and (end_date is null or end_date > current_date)")
```

According to pre-prod, there are >2000 personal circumstances with future end dates.

projects/unpaid-work-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/personaldetails/entity/PersonalCircumstanceEntity.kt
```diff
@@ -0,0 +1,78 @@
...
+@Table(name = "personal_circumstance")
+class PersonalCircumstanceEntity(
+    @Id @Column(name = "personal_circumstance_id")
+    var id: Long,
```
marcus-bcl, line 21: Does this need to be a `var`?  Same for a couple of other id fields

projects/unpaid-work-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/personaldetails/entity/PersonalContactEntity.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    @Column(name = "surname")
+    val surname: String,
+
+    @Column(name = "mobile_number")
+    val mobileNumber: String?,
+
+    @ManyToOne
+    @JoinColumn(name = "address_id")
+    val address: AddressEntity?,
```
marcus-bcl, line 42: Do we need to check for soft deleted addresses here?  I guess it must be unlikely that the personal contact is active but the address is soft deleted though...

marcus-bcl (approved):
Looks good, few minor comments.

projects/unpaid-work-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/personaldetails/entity/PersonalCircumstanceEntity.kt
marcus-bcl: Looks like this is still a var @stevomcallister ?

PR #751, opened 2023-01-24
PI-732 move person search indexer to CP

anthony-britton-moj:


projects/person-search-index-from-delius/deploy/values.yaml
```diff
@@ -0,0 +1,46 @@
+generic-service:
+  nameOverride: person-search-index-from-delius
+  replicaCount: 1
+  resources:
+    limits:
+     cpu: 2
+     memory: 2GB
+    requests:
+    cpu: 1
+    memory: 1GB
```
marcus-bcl, line 10:
```suggestion
  resources:
    limits:
      cpu: 2
      memory: 2GB
    requests:
      cpu: 1
      memory: 1GB
```
This should fix the indentation

projects/person-search-index-from-delius/deploy/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+  readinessProbe:
+    httpGet:
+      path: localhost:9600
```
marcus-bcl, line 14: By setting the port in the `image` block, I guess the path can change to just `/`
```suggestion
      path: /
```

projects/person-search-index-from-delius/deploy/values.yaml
```diff
@@ -0,0 +1,46 @@
...
+  replicaCount: 1
+  resources:
+    limits:
+     cpu: 2
+     memory: 2GB
+    requests:
+    cpu: 1
+    memory: 1GB
+
+  readinessProbe:
```
marcus-bcl, line 12: Do you need another block for livenessProbe?

PR #774, opened 2023-01-26
PI-732 access via namespace

anthony-britton-moj:


stevomcallister (approved):

marcus-bcl (approved):
Maybe worth setting [continue-on-error: true](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) to the Sonar step in future if it keeps failing...

PR #796, opened 2023-01-31
PI-724 - case details endpoint

stevomcallister:
…ntegration-services into main


marcus-bcl (commented):
Code looks good, but no unit tests?

stevomcallister:
> Code looks good, but no unit tests?

Don't really see the point for this one as there isn't any real logic and the integration tests cover almost all of it.

PR #805, opened 2023-02-02
PI-770

anthony-britton-moj:


marcus-bcl (approved):
Nice - much prefer the use of ApplicationListener over CommandLineRunner.

Although just noticed ktlint is failing on the templates ☹️ 
```
/home/runner/work/hmpps-probation-integration-services/hmpps-probation-integration-services/projects/test-project/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/DataLoader.kt:3:1 Imports must be ordered in lexicographic order without any empty lines in-between with "java", "javax", "kotlin" and aliases in the end
```

marcus-bcl (approved):

PR #818, opened 2023-02-03
PI-762 jib cache

anthony-britton-moj:


buildSrc/src/main/kotlin/uk/gov/justice/digital/hmpps/plugins/JibConfigPlugin.kt
```diff
@@ -34,23 +38,33 @@ class JibConfigPlugin : Plugin<Project> {
...
-                dependsOn(copyAgentTask)
+            val assemble = project.tasks.named("assemble")
+            project.tasks.withType<BuildImageTask>().named("jib") {
+                doFirst {
+                    jib!!.to.tags = setOf("${project.version}")
+                }
+                dependsOn(copyAgent, `copyAppInsightsConfig`, assemble)
+                inputs.dir("deploy")
+                inputs.dir("src")
+                inputs.file("build.gradle.kts")
```
marcus-bcl, line 65: Could we pick this up again today?  It occurred to me that this won't build if any of the libs or dependency versions change.

marcus-bcl (approved):

PR #846, opened 2023-02-13
PI-793 person endpoint for crn and noms

anthony-britton-moj:


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/resource/PersonResource.kt
```diff
@@ -0,0 +1,26 @@
...
+
+@RestController
+@RequestMapping("/person")
+class PersonResource(private val personService: PersonService) {
+
+    @PreAuthorize("hasRole('ROLE_ALLOCATION_CONTEXT')")
+    @GetMapping("/{value}")
+    fun findPerson(
+        @PathVariable value: String,
+        @RequestParam(required = false, defaultValue = "CRN") type: IdentifierType
```
marcus-bcl, line 20: Should we think about how we want to handle resources with multiple IDs going forward?  Community API uses a path part (e.g. `/person/crn/A123456`), is there a reason for switching to a request parameter here?

I've got no preference either way, but it'd be great to agree on a consistent pattern for these.

A couple of other possible approaches for comparison:
* Named path part: `/person/crn=A123456` and `/person/noms=A1234BC`,
* Request parameter as filter: `/persons?crn=A123456` and `/persons?noms=A1234BC`.

marcus-bcl (commented):
Looks good, nice that we can re-use the case type query.  One query on the URL, but nothing major - just a check that we're all happy with the pattern.

stevomcallister (approved):

PR #861, opened 2023-02-15
PI-813: Pre-Sentence Report AsyncAPI spec

mxco86:


projects/pre-sentence-reports-to-delius/tech-docs/source/async-api-reference.html.md.erb
```diff
@@ -0,0 +1,15 @@
...
+weight: 30
+---
+
+<header id="api" class="anchored-heading govuk-heading-xl" style="padding-top: 30px">
+    <a href="#api" class="anchored-heading__icon" aria-hidden="true" tabindex="-1"></a>
+    Async API Reference
+</header>
+<style>.technical-documentation > h1 { display: none !important }</style>
+
+<h1>ASyncAPI</h1>
```
marcus-bcl, line 13: You don't need this bit, it was a hack to get around the tech-docs OpenAPI parser messing with the header.  You should be able to just use Markdown as normal.

```suggestion
# Async API Reference
```

marcus-bcl (approved):

PR #871, opened 2023-02-17
PI-817 recall from temporary licence

anthony-britton-moj:


stevomcallister (approved):

marcus-bcl:
> Andrew Logan and James both seem to be asking for it to be reverted. Andrew actually stated `I'm happy for that to be implemented` and James was one of the people on the call when the original ticket was reverted. If they are both now saying that we should undo the reversion and return to this behaviour, it is unlikely they will change their mind again - until they see the results. It seems like, based on the cases Karen has looked at, the behaviour would be correct - so maybe their original concern is an edge case rather than the norm.

Let's wait for a response from Karen before merging anything, there's no need to rush from our side.

PR #887, opened 2023-02-21
PI-823 Improve deploy process to use gradle tasks to determine what to deploy

anthony-britton-moj:


projects/person-search-index-from-delius/build.gradle.kts
```diff
@@ -6,4 +6,12 @@ tasks {
         tag("latest", "ghcr.io/ministryofjustice/hmpps-probation-integration-services/${project.name}:latest")
         files("container")
     }
+    named("docker") {
+        doLast {
+            val dir = File("${project.parent?.projectDir}/changed")
```
marcus-bcl, line 11: Would this work for nested projects?  E.g. for the person-search management image

.github/actions/build-and-test/action.yml
```diff
@@ -67,6 +65,15 @@ runs:
...
       run: ./gradlew docker
 
+    - name: Check Changed Images
+      id: changed-images
+      shell: bash
+      run: |
+        echo "Changed Projects: $(ls changed)"
+        for file in $(ls changed); do projects=(${projects[@]} "$file"); done; \
+        projects_array=$(jq --compact-output --null-input '$ARGS.positional' --args -- ${projects[@]}) \
+        && echo "projects=${projects_array}" >> "$GITHUB_OUTPUT"
```
marcus-bcl, line 75: Could this be simplified to:
```suggestion
      run: echo "projects=$(ls changed | jq --raw-input . | jq --slurp --compact-output .)" >> $GITHUB_OUTPUT
```

projects/person-search-index-from-delius/build.gradle.kts
marcus-bcl: Could you use `${project.rootDir}/changed` ?

buildSrc/src/main/kotlin/uk/gov/justice/digital/hmpps/plugins/JibConfigPlugin.kt
```diff
@@ -57,6 +58,12 @@ class JibConfigPlugin : Plugin<Project> {
                         }
                     }
                 }
+                doLast {
+                    val dir = File("${project.parent?.projectDir}/changed")
```
marcus-bcl, line 62: Looks good.  Is it worth using `rootDir` here too for consistency?

marcus-bcl (approved):

PR #902, opened 2023-02-22
PI-836 Fix audited Interactions not recorded in delius

anthony-britton-moj:


stevomcallister (approved):

libs/commons/src/main/kotlin/uk/gov/justice/digital/hmpps/logging/LazyLogger.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.logging
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+fun <T : Any> T.logger(): Lazy<Logger> {
```
marcus-bcl, line 6: Do we need this?  It doesn't look like we're using it anymore?

marcus-bcl (commented):
Code looks good, but not sure I can tell what the problem was? 😄

marcus-bcl:
> > Code looks good, but not sure I can tell what the problem was? 😄
> 
> Basically we were using the SecurityContext -- provided by Spring Boot. However, when processing messages, that doesn't actually get set. Therefore, the ContextRunnable we have the copies the SecurityContext and RequestContext (from the calling thread to the async thread) had nothing to copy. We were also always checking that the principal was a ServicePrincipal - otherwise not carrying out auditing either. Of course, the only time we have an ServicePrincipal is from the ServiceContext -- this is the bit that is set on app start up that prevents the app from starting if the db user doesn't exist. Essentially, this can never not have a user now anyway. Thus the changing of some nullable user ids to non-nullable

Okay, think I understand.  Is that not what the `DelegatingSecurityContextRunnable` is for?  I guess implementing our own gives us more control... but not sure if we're reinventing the wheel?

PR #937, opened 2023-02-27
PI-851 Add Tier Calculation Required into OGRS Consumer

stevomcallister:


marcus-bcl (approved):

projects/risk-assessment-scores-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/RiskAssessmentService.kt
```diff
@@ -69,9 +76,21 @@ class RiskAssessmentService(
                 )
             )
             createContact(person, event, assessmentDate, ogrsScore)
+            createManagementTierEvent(person)
```
marcus-bcl, line 79: Already approved but just noticed - do we need to do this for update as well?

stevomcallister: ha! yes

marcus-bcl (approved):

PR #944, opened 2023-03-01
PI-869 Create error on terminated event number

stevomcallister:


projects/risk-assessment-scores-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/RiskAssessmentService.kt
```diff
@@ -52,6 +52,10 @@ class RiskAssessmentService(
         // validate that the offender has an event with this event number
         val event = eventRepository.getByCrn(crn, eventNumber.toString())
 
+        if (!event.active) {
+            throw ConflictException("[crn=${person.crn}][eventNumber=$eventNumber]: Event number is Terminated")
```
marcus-bcl, line 56: Don't know if we need the `[crn=${person.crn}][eventNumber=$eventNumber]: ` prefix bit in our code.  I think they just included it in the story as that's what Yuri prefixes the error messages coming out of the RSR stored procedure with.

It can't hurt though, so happy to leave it.

marcus-bcl (dismissed):

projects/risk-assessment-scores-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/RiskAssessmentService.kt
marcus-bcl: Also, is there a difference between terminated and inactive for an event?  Or do they mean the same thing?

stevomcallister: ah ok - I really don't like it either but wondered if the testers might be picky about it. I'll change it to a text description.

stevomcallister: I was led to believe that inactive and terminated were the same thing. There isn't really a terminated state on an event - its the disposal that gets terminated but rather than add disposal to the code it was easier to use the active status.

marcus-bcl (approved):

PR #984, opened 2023-03-08
PI-878 

anthony-britton-moj:
Allow person-search-index-from-delius docker build/push to be cached

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -70,18 +70,18 @@ subprojects {
     apply { plugin(ClassPathPlugin::class.java) }
     apply { plugin("org.sonarqube") }
     apply { plugin("org.jetbrains.kotlin.kapt") }
-}
 
-tasks {
-    withType<BootRun> {
-        if (System.getProperty("spring.profiles.active", System.getenv("SPRING_PROFILES_ACTIVE")) == "dev") {
-            classpath = sourceSets.getByName("dev").runtimeClasspath
+    tasks {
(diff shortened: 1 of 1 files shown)


marcus-bcl (approved):
I wonder if at this point, given the Gradle Docker plugin is deprecated, we should just use Docker to build this project instead of Gradle?  

The Docker actions (https://github.com/docker/build-push-action) support using the GitHub Actions cache as a layer cache as well.  We'd just need to think of a way of detecting if there were changes to deploy, I guess if the image tag hasn't changed then it will have used the layer cache and we wouldn't need to deploy anything?

marcus-bcl:
> > I wonder if at this point, given the Gradle Docker plugin is deprecated, we should just use Docker to build this project instead of Gradle?
> > The Docker actions ([docker/build-push-action](https://github.com/docker/build-push-action)) support using the GitHub Actions cache as a layer cache as well. We'd just need to think of a way of detecting if there were changes to deploy, I guess if the image tag hasn't changed then it will have used the layer cache and we wouldn't need to deploy anything?
> 
> Yeah, using grade to build a docker file was always a bit weird. The image tag though is based on the version, so changes on each run, this why it is built and deployed on each merge previous to this PR. Worth considering though definitely. Potentially something based on container dir changes and sha value of base image?

Sorry I meant sha not tag 😄  but yeah we could also fall back to the file-based change detection for this project.  I'll stick a ticket on the backlog.

PR #1017, opened 2023-03-10
Update mutation strategy to not use temporary tables

anthony-britton-moj:
see https://github.com/hibernate/hibernate-orm/blob/main/documentation/src/main/asciidoc/userguide/chapters/batch/Batching.adoc#non-temporary-table-bulk-mutation-strategies


marcus-bcl:
Is this causing us a problem?  As far as I know we don't have permissions to create temporary tables anyway, so I thought we'd have seen an error?

izzyL-DJ (approved):

marcus-bcl:
> > Is this causing us a problem? As far as I know we don't have permissions to create temporary tables anyway, so I thought we'd have seen an error?
> 
> Have you added permissions to create the temporary tables now? We didn't have and we were getting errors because of not having permissions -- thus why we added the first property.

No, we have the same permissions as before.  So I'm wondering if this PR is needed?

PR #1030, opened 2023-03-14
PI-895 - Add allocation detail endpoint

anthony-britton-moj:


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AllocationDemandService.kt
```diff
@@ -134,6 +136,23 @@ class AllocationDemandService(
...
     }
 
+    fun getDetails(requests: AllocationDetailRequests): AllocationDetails {
+        val cases = personRepository.findAllByCrnAndSoftDeletedFalse(requests.cases.map { it.crn })
+            .associateBy { it.crn }
+        val staff = staffRepository.findAllByCodeIn(requests.cases.map { it.staffCode })
+            .map { it.toStaffMember() }
+            .associateBy { it.code }
+        return AllocationDetails(
+            requests.cases.map {
```
marcus-bcl, line 146: Could you use `mapNotNull` here, to save making the crn and name nullable?

marcus-bcl: Makes sense, so they might send in a CRN/Staff pair but one or the other might not exist.

marcus-bcl (approved):

PR #1132, opened 2023-03-20
PI-914 - Referral Started

anthony-britton-moj:


projects/refer-and-monitor-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/NsiService.kt
```diff
@@ -30,14 +35,47 @@ class NsiService(
...
+    @Transactional
+    fun startNsi(crn: String, rs: ReferralStarted) = audit(MANAGE_NSI) { audit ->
+        val nsi = nsiRepository.findByPersonCrnAndExternalReference(crn, rs.urn)
+            ?: createNsi.new(crn, rs).apply {
+                statusHistoryRepository.save(statusHistory())
+                contactRepository.save(contact(NSI_REFERRAL.value, referralDate))
+                contactRepository.save(statusChangeContact())
+                contactRepository.save(contact(NSI_COMMENCED.value, actualStartDate!!))
+                audit["offenderId"] = person.id
+                audit["nsiId"] = id
```
marcus-bcl, line 51: Is it worth adding these to the audit context even if we're updated an existing NSI?

marcus-bcl (approved):

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #1133, opened 2023-03-21
PI-914 - add table name for provider

anthony-britton-moj:


marcus-bcl:
Do you also need to update https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/projects/refer-and-monitor-and-delius/deploy/database/access.yml for accessing nsi_manager?

stevomcallister (dismissed):

stevomcallister (approved):

PR #1237, opened 2023-03-28
PI-912 PI-926 updated for transfer for RoL and RoTL

anthony-britton-moj:


projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/recall/RecallService.kt
```diff
@@ -57,9 +58,12 @@ val EOTL_RECALL_CONTACT_NOTES = """${System.lineSeparator()}
     |The date may reflect an update after the date the actual Recall/Return to Custody occurred
 """.trimMargin()
 
+const val FEATURE_FLAG_TRANSFERRED = "messages.recall.transferred"
+
 @Service
 class RecallService(
     auditedInteractionService: AuditedInteractionService,
+    private val featureFlags: FeatureFlags,
```
marcus-bcl, line 66: You'll need to add the config to values-prod.yml too.  See https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/bd9a93c86112cd8a3717c9cc8631644d27cf34a7/projects/risk-assessment-scores-to-delius/deploy/values-prod.yml#L7-L11

marcus-bcl: Good to know - any reason for the `messages_received_` prefix?  I wonder if we should prefix them with the service name or something like that?

marcus-bcl (approved):

PR #1267, opened 2023-04-04
PI-963 Added responsible officer endpoint

anthony-britton-moj:


projects/refer-and-monitor-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/resource/ProbationCaseResource.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import uk.gov.justice.digital.hmpps.api.model.ResponsibleOfficer
+import uk.gov.justice.digital.hmpps.service.ManagerService
+
+@RestController
+@RequestMapping("probation-case/{crn}")
+class ProbationCaseResource(private val managerService: ManagerService) {
+    @PreAuthorize("hasRole('CRS_RESPONSIBLE_OFFICER')")
```
marcus-bcl, line 14: Do we need a separate role for this?  Will it always be called by the same client (refer-and-monitor)?

projects/refer-and-monitor-and-delius/build.gradle.kts
```diff
@@ -14,14 +17,17 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-security")
     implementation("org.springframework.boot:spring-boot-starter-validation")
     implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("org.springframework.boot:spring-boot-starter-data-ldap")
```
marcus-bcl, line 20: You'll need to update the values.yml file too with the LDAP URL and credentials.

marcus-bcl (commented):
Looks good

marcus-bcl (approved):

PR #1279, opened 2023-04-06
PI-971 - updated to handle main address correctly

stevomcallister:


projects/unpaid-work-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/casedetails/entity/CaseEntity.kt
```diff
@@ -52,8 +52,15 @@ class CaseEntity(
...
     val mobileNumber: String? = null,
-    @OneToOne(mappedBy = "person")
-    val mainAddress: CaseAddress? = null,
+    @OneToMany(mappedBy = "person")
+    @Where(clause = """
+        address_status_id = (
+        select at.standard_reference_list_id from r_standard_reference_list at 
+        join r_reference_data_master rm on rm.reference_data_master_id = at.reference_data_master_id
+        where rm.code_set_name = 'ADDRESS STATUS' and at.code_value = 'M'
+        )
```
marcus-bcl, line 61: Nice! Yeah that's what I was talking about with the sub-select in the where clause.  You can probably put it back to being a one-to-one now as I'm sure they can only have one active main address?

marcus-bcl: @stevomcallister just checked and yes you can make it a one-to-one:
```sql
select offender_id 
from offender_address 
where address_status_id=521
and soft_deleted = 0
group by offender_id
having count(*) > 1;

> 0 results
```

stevomcallister: ok I'll update it.

marcus-bcl (approved):

PR #1421, opened 2023-04-19
PI-961

stevomcallister:


projects/effective-proposal-framework-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/epf/entity/PrisonManager.kt
```diff
@@ -0,0 +1,30 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import org.hibernate.annotations.Where
+import org.springframework.data.annotation.Immutable
+
+@Entity
+@Table(name = "prison_offender_manager")
+@Where(clause = "soft_deleted = 0 and active = 1")
```
marcus-bcl, line 14: Should this be active_flag?

```suggestion
@Where(clause = "soft_deleted = 0 and active_flag = 1")
```

marcus-bcl (commented):
Looks good.

There are a few issues from bootstrapping still to do btw: https://github.com/ministryofjustice/hmpps-probation-integration-services/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+effective-proposal-framework-and-delius

Should probably also update the .trivyignore to match the other projects.

PR #1432, opened 2023-04-21
PI-1016 Scale down pods out-of-hours in dev/preprod

marcus-bcl:
Helm chart changes were released in 2.6.0: https://github.com/ministryofjustice/hmpps-helm-charts/pull/91


marcus-bcl:
> Should we let people know that preprod services will be shut down ? Some teams may leave their services running in pre-prod?

Could do, although I can't see any activity out-of-hours in preprod so I think we're fine.

PR #1545, opened 2023-05-12
PI-1082 Fix potential duplicate appointment creation

marcus-bcl:
Previously we were selecting the NSI for update, but not making any changes to it. This meant parallel requests could succeed in creating contacts with the same external_reference.


stevomcallister (approved):

marcus-bcl:
> This isn't necessary . We shouldn't be updating the NSI just for this purpose, that's the whole point in select for update

I thought the same thing, but the test consistently fails and creates duplicates without the change - however it consistently passes after the change.  Maybe a H2 thing?  I'll keep playing with it, to see if I can get it to pass without updating the NSI.

marcus-bcl:
The test passes against Oracle, so closing this in favour of https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/1546

PR #1547, opened 2023-05-15
PI-1073 Probation area endpoint

stevomcallister:


projects/pathfinder-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationAreaEntity.kt
```diff
@@ -0,0 +1,102 @@
...
+        """
+        select new uk.gov.justice.digital.hmpps.entity.ProbationAreaDistrict(pa.code, pa.description, d.code, d.description)
+        from ProbationAreaEntity pa 
+        join Borough b on b.probationAreaId = pa.id
+        join District d on d.borough.id = b.id
+        where pa.description not like 'ZZ%'
+        and d.code <> '-1'
+        and pa.selectable = true
+        and d.selectable = true
+        and b.selectable = true
```
marcus-bcl, line 96: Should these be `@Where` annotations on the entities?

projects/pathfinder-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProbationAreaService.kt
```diff
@@ -0,0 +1,29 @@
...
+import uk.gov.justice.digital.hmpps.model.ProbationAreaContainer
+
+@Service
+class ProbationAreaService(private val probationAreaRepository: ProbationAreaRepository) {
+    fun getProbationAreas(): ProbationAreaContainer {
+        val pas = mutableListOf<ProbationArea>()
+        val pad = probationAreaRepository.probationAreaDistricts()
+
+        // create a set of probation areas
+        val paCodes: Set<Pair<String, String>> = pad.map { Pair(it.pCode, it.pDesc) }.toSet()
```
marcus-bcl, line 17: Is this `toSet()` call required to remove duplicates from the cross-join?  If so I guess changing `districts` and `boroughs` to `Set` instead of `List` on the entities would have the same effect - not sure if that would simplify things?

marcus-bcl (approved):
Approved with a couple of comments

projects/pathfinder-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProbationAreaService.kt
stevomcallister: The set is for the probation areas so I think this is the only way to do it.

projects/pathfinder-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationAreaEntity.kt
stevomcallister: tried this and it doesn't work.

PR #1556, opened 2023-05-17
PI-1040: Document make recall decisions and delius

mxco86:


projects/make-recall-decisions-and-delius/tech-docs/source/img/make-recall-decisions-and-delius-context-map.png
marcus-bcl, line 1: Does the bottom diagram indicate that we update Offender/Offender Manager etc as part of the recall consideration process?  We don't actually update any of that, we only create contacts: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/46736b8adbb9c65fcdec537d1ebf4a39f32c284d/projects/make-recall-decisions-and-delius/deploy/database/access.yml#L5-L7

marcus-bcl: No worries, makes sense - I've approved it anyway for now if you want to merge as is

marcus-bcl (approved):

PR #1557, opened 2023-05-17
PI-1075 added convictions endpoint

stevomcallister:


projects/pathfinder-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ConvictionService.kt
```diff
@@ -0,0 +1,33 @@
...
+        }
+        val convictionModels = mutableListOf<Conviction>()
+        convictions.map { convictionEventEntity ->
+            val offences = mutableListOf<Offence>()
+            offences.add(Offence(convictionEventEntity.mainOffence?.offence?.description ?: "unknown", true))
+            convictionEventEntity.additionalOffences.forEach { offences.add(Offence(it.offence.description, false)) }
+            convictionModels.add(
+                Conviction(
+                    convictionEventEntity.convictionDate,
+                    convictionEventEntity.disposal?.type?.description ?: "unknown",
```
marcus-bcl, line 25: Is it worth just making this nullable?

marcus-bcl (approved):

projects/pathfinder-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ConvictionService.kt
stevomcallister: ok I'll fix this in the next PR

PR #1652, opened 2023-06-01
PI-1134 Handle null attended flag

anthony-britton-moj:


marcus-bcl (commented):
Does this not just move the error to an IllegalStateException instead of a DecodeException?

marcus-bcl (approved):

PR #1727, opened 2023-06-09
PI-1137 adding criminogenic needs endpoint

stevomcallister:


projects/sentence-plan-and-oasys/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/ApiController.kt
```diff
@@ -7,11 +7,24 @@ import org.springframework.web.bind.annotation.RestController
...
+        return Needs(
+            listOf(
+                Need("Accommodation"),
+                Need("Education/Employability"),
+                Need("Relationships"),
+                Need("Lifestyle"),
+                Need("Drug misuse"),
+                Need("Alcohol misuse"),
+                Need("Thinking & Behaviour"),
+                Need("Attitudes")
```
marcus-bcl, line 26: Would be good to have a key for each of these too if possible, this is how we reference them in the UI: https://github.com/ministryofjustice/hmpps-sentence-plan-ui/pull/13/files#diff-4123d483853f30d2316283afb008fd49932171d4ad6650e3ba64d1f8d72cc683R50-R57

So for example: `key = "employability", description = "Education, training and employment"`

marcus-bcl (approved):

PR #1733, opened 2023-06-12
PI-1138 release management

anthony-britton-moj:


projects/approved-premises-and-delius/.trivyignore
```diff
@@ -1,6 +1,9 @@
...
 # Suppressed as this service doesn't handle file uploads
-CVE-2023-24998 exp:2023-06-01
+CVE-2023-24998 exp:2023-07-01
 
 # Suppressed as we do not process any untrusted YML content
-CVE-2022-1471 exp:2023-06-01
+CVE-2022-1471 exp:2023-07-01
+
+# Suppressed because we do not process any untrusted XML content
+CVE-2023-2650 exp:2023-07-01
```
marcus-bcl, line 9: Don't think this is to do with XML is it?  It's a vulnerability in OpenSSL which does affect us if we're making HTTPS calls.  The fix is in libssl 3.1.1, which should be added to the base image very soon according to https://github.com/alpinelinux/docker-alpine/issues/328#issuecomment-1586315306

Still, makes sense to ignore it for now - but maybe set a shorter expiry given we're expecting this to be fixed soon?

marcus-bcl (approved):

PR #1816, opened 2023-06-22
DST-14800 Disable publishing while NDST perform testing

marcus-bcl:


marcus-bcl:
Testing has finished (I scaled the service down manually).

marcus-bcl:
Delius team are testing again.

PR #1825, opened 2023-06-23
PI-1191 Disable index auto-creation

marcus-bcl:
This configures the cluster settings any time a container starts, to ensure the settings are always up to date.


marcus-bcl:
> Is there a reason we don't just turn off automatic index creation? That way the a new index will not be created except for when the script is ran to create a new one

Other than I didn't think of it, no 😄 

I'll update the PR to disable auto index creation instead.

PR #1846, opened 2023-06-27
PI-1225 updated logic to set outcome on future appointments

anthony-britton-moj:
instead of deleting


marcus-bcl (approved):
Do we also need to withdraw linked contacts?

marcus-bcl:
> > Do we also need to withdraw linked contacts?
> 
> We can discuss it on the standup, but I would say not for the following: Linked contacts are usually enforcement actions - which means at some point someone has marked it as failed to attend, then updated it to be in the future, presumably to attempt to `delete` it (or at least remove it from visibility). If we don't get the end referral message that linked contact would remain there anyway and visible. Only the appointment would be removed from visibility.
> 
> It is an edge case, doesn't happen often, as we can see from the Unable to delete future contact exceptions 26 over 30 days compared to the 120K appointments updated. Most of these (17) were on the same day. Likely, the offender manager has already seen the enforcement action and deemed it not worthy of action. I cannot image why they would update the contact if they deemed it to be worthy of action. Then it's a matter of educating the users updating it to ridiculous dates in the future that this isn't the correct way of dealing with it.

Yes makes sense, like you say I'm pretty sure the only ones we've seen are where they've tried to update the date far into the future instead of deleting - which is a Delius problem, not ours.

PR #1855, opened 2023-06-30
PI-1181 Prevent Hibernate Creating Global Temp Tables

anthony-britton-moj:
based on this:
https://hibernate.atlassian.net/browse/HHH-15550 

we just need to set the creation of these tables on startup to false


marcus-bcl (commented):
Nice. Does it need adding to the templates too?

marcus-bcl (approved):

PR #1970, opened 2023-07-19
PI-1305 Reduce dependencies in app insights

marcus-bcl:


stevomcallister (approved):
I'm sure its great - have a tick

PR #2012, opened 2023-07-26
PI-1340 remove RO updates and some refactoring

anthony-britton-moj:


projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/event/entity/Event.kt
```diff
@@ -0,0 +1,67 @@
...
+import jakarta.persistence.OneToOne
+import jakarta.persistence.Version
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Modifying
+import org.springframework.data.jpa.repository.Query
+import uk.gov.justice.digital.hmpps.integrations.delius.person.entity.Person
+import java.time.ZonedDateTime
+
+@Entity
+class Event(
```
stevomcallister, line 17: @where annotation for softDeleted and action?

PR #2030, opened 2023-07-31
PI-1354 add handling of identifier added / updated events

anthony-britton-moj:


projects/prison-custody-status-to-delius/src/main/resources/application.yml
```diff
@@ -16,9 +16,33 @@ spring:
           global_temporary:
             create_tables: false
             drop_tables: false
+  security.oauth2.client:
+    registration:
+      prison-custody-status-to-delius:
+        provider: hmpps-auth
+        authorization-grant-type: client_credentials
+        client-id: ${oauth2.client-id}
+        client-secret: ${oauth2.client-secret}
```
marcus-bcl, line 25: Do we need to request new HMPPS Auth clients for this, and reference the credentials in the values.yaml file?

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -5,52 +5,80 @@ import uk.gov.justice.digital.hmpps.converter.NotificationConverter
...
+        when (prisonerMovement) {
+            is PrisonerMovement.Received -> {
+                val outcome = recallService.recall(prisonerMovement)
+                telemetryService.trackEvent(outcome.name, prisonerMovement.telemetryProperties())
+            }
+
+            is PrisonerMovement.Released -> {
+                telemetryService.trackEvent("IdentifierAddedForReleasedPrisoner", prisonerMovement.telemetryProperties())
+                throw IllegalStateException("NomsNumberAddedAfterRelease")
+            }
```
marcus-bcl, line 81: Shouldn't this only be thrown if they're currently in custody in Delius?

projects/prison-custody-status-to-delius/deploy/values.yaml
```diff
@@ -11,6 +11,9 @@ generic-service:
   namespace_secrets:
     common:
       SPRING_DATASOURCE_URL: DB_URL
+    prison-custody-status-to-delius-client-credentials:
```
marcus-bcl, line 14: You'll also need to add the token url to dev/preprod/prod e.g. 
```yaml
SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_HMPPS-AUTH_TOKEN-URI: https://sign-in-dev.hmpps.service.justice.gov.uk/auth/oauth/token
```

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
marcus-bcl: As it is, this will throw an error if someone is matched while in the community.  Which could happen quite often?

marcus-bcl: Happy to have a quick chat about it, I may be missing something - it feels like we could continue with a null/unknown prison, then rely on the validation method to log an UnexpectedInstitution event as normal.  We wouldn't get alerted for the exception, but not the end of the world.

marcus-bcl (approved):

PR #2295, opened 2023-09-20
PI-1473 add enhancement ability for domain events

anthony-britton-moj:
Part one about ready - ability to add the detail url (or any other information desired) to a domain event
Part two to add the api reading ability - detail url added for probation-case.engagement.created


stevomcallister (approved):

marcus-bcl (approved):
Just a thought, happy to leave it as is - but should we prefix the path with something like "/details/..."?  E.g.
```
domain-events-and-delius.hmpps.service.justice.gov.uk/details/probation-case.engagement.created/A123456
```
in case we ever need to add any other API endpoints (unrelated to event details) to this service?

PR #2300, opened 2023-09-21
PI-1478 changes to retry exceptions

stevomcallister:


libs/messaging/src/main/kotlin/uk/gov/justice/digital/hmpps/listener/AwsNotificationListener.kt
```diff
@@ -25,7 +31,16 @@ class AwsNotificationListener(
...
         try {
-            handler.handle(message)
+            retry(
+                3,
+                listOf(
+                    FeignException.NotFound::class,
+                    RetryableException::class,
+                    ObjectOptimisticLockingFailureException::class,
+                    CannotCreateTransactionException::class,
+                    CannotGetJdbcConnectionException::class
```
marcus-bcl, line 41: Is it worth including `CannotAcquireLockException` here too?

libs/commons/src/main/kotlin/uk/gov/justice/digital/hmpps/retry/Retry.kt
```diff
@@ -1,13 +1,24 @@
 package uk.gov.justice.digital.hmpps.retry
 
-fun <T> retry(maxRetries: Int, code: () -> T): T {
-    var throwable: Throwable? = null
-    (1..maxRetries).forEach { _ ->
+import kotlin.reflect.KClass
+
+fun <T> retry(maxRetries: Int, exceptions: List<KClass<out Exception>> = listOf(Exception::class), code: () -> T): T {
+    var throwable: Throwable?
+    (1..maxRetries).forEach { count ->
```
marcus-bcl, line 7: Do we need any kind of delay or back-off between retries?

libs/commons/src/main/kotlin/uk/gov/justice/digital/hmpps/retry/Retry.kt
```diff
@@ -1,13 +1,24 @@
...
+            val matchedException = exceptions.firstOrNull { it.isInstance(e) }
+            throwable = if (matchedException != null && count < maxRetries) {
+                null
+            } else {
+                e
+            }
+
+            if (throwable != null) {
+                throw throwable!!
+            }
```
marcus-bcl, line 20: I'm a bit confused by this... probably just me... where do we return the successful value outside the forEach?

libs/messaging/src/main/kotlin/uk/gov/justice/digital/hmpps/listener/AwsNotificationListener.kt
stevomcallister: ok

libs/commons/src/main/kotlin/uk/gov/justice/digital/hmpps/retry/Retry.kt
stevomcallister: added a delay and using Feign retry stuff

PR #2377, opened 2023-10-06
PI-1542

anthony-britton-moj:


projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/ProbationCase.kt
```diff
@@ -115,15 +115,34 @@ class CommunityManagerTeam(
...
+
     @Id
     @Column(name = "team_id")
     val id: Long
 )
 
+@Immutable
+@Entity
+@Table(name = "district")
+class Ldu(
```
marcus-bcl, line 130: Is LDU the right name for this?  I thought it was called "Local Admin Unit" now.

stevomcallister (approved):

projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/ProbationCase.kt
marcus-bcl: No worries, I think actually Consider a Recall refer to it as LDU too - so I guess we can treat LDU/LAU as interchangeable.

marcus-bcl (approved):

PR #2421, opened 2023-10-19
PI-1505 Initial grouped documents endpoint

marcus-bcl:


marcus-bcl:
> Approved - however I would question why the entities are `data` classes ? Doesn't seem like they need to be and given the potential complications of data classes with JPA it might be better to avoid them

Good point - updated

PR #2433, opened 2023-10-23
PI-1579 CAS3 Referral submitted processing

stevomcallister:


projects/cas3-and-delius/deploy/values.yaml
```diff
@@ -12,6 +11,9 @@ generic-service:
   namespace_secrets:
     common:
       SPRING_DATASOURCE_URL: DB_URL
+    as3-and-delius-client-credentials:
```
marcus-bcl, line 14:
```suggestion
    cas3-and-delius-client-credentials:
```

projects/cas3-and-delius/deploy/values-dev.yml
```diff
@@ -9,6 +7,9 @@ generic-service:
 
   env:
     SENTRY_ENVIRONMENT: dev
+    SPRING_SECURITY_OAUTH2_CLIENT_PROVIDER_HMPPS-AUTH_TOKEN-URI: https://sign-in-dev.hmpps.service.justice.gov.uk/auth/oauth/token
```
marcus-bcl, line 10: Do we need to add the approved-premises-api URL in here too?

projects/cas3-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/Contact.kt
```diff
@@ -0,0 +1,125 @@
...
+    val type: ContactType,
+
+    @Lob
+    var notes: String,
+
+    @Column(name = "contact_date")
+    var date: LocalDate,
+
+    @Column(name = "contact_start_time")
+    var startTime: ZonedDateTime,
```
marcus-bcl, line 51: Few vars in here, can they be vals?

projects/cas3-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/PersonManager.kt
```diff
@@ -0,0 +1,50 @@
...
+    val staffEmployeeId: Long,
+
+    @Column(name = "probation_area_id")
+    val probationAreaId: Long,
+
+    @Column(name = "active_flag", columnDefinition = "NUMBER")
+    val active: Boolean = true,
+
+    @Column(name = "soft_deleted", columnDefinition = "NUMBER", nullable = false)
+    var softDeleted: Boolean = false
```
marcus-bcl, line 36: Same here, can this be a val?

projects/cas3-and-delius/deploy/database/access.yml
```diff
@@ -2,8 +2,11 @@ database:
...
     username_key: /cas3-and-delius/db-username
     password_key: /cas3-and-delius/db-password
+    tables:
+      - audited_interaction
+      - contact
 
   audit:
     username: Cas3AndDelius
-    forename: Probation Integration # TODO change this to something meaningful for your service
+    forename: cas3
```
marcus-bcl, line 11: This is the user-facing name on the contact, will user's know it by CAS3?  Or should we call it something like "Transitional Accommodation (CAS3)"

marcus-bcl (approved):
Few very minor things, but looks good

projects/cas3-and-delius/deploy/values-dev.yml
stevomcallister: I didn't think so since we only use the url from the message?

marcus-bcl: Ah good point sorry!

PR #2445, opened 2023-10-25
PI-1586 add contacts for POM Allocation and RO change

anthony-britton-moj:


projects/manage-pom-cases-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/contact/entity/Contact.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    @Column(name = "offender_id")
+    val personId: Long,
+
+    @ManyToOne
+    @JoinColumn(name = "contact_type_id")
+    val type: ContactType,
+
+    @Column(name = "contact_date")
+    var date: LocalDate,
```
stevomcallister, line 39: var? doesn't look like you are changing this later

stevomcallister (dismissed):

stevomcallister (approved):

PR #2452, opened 2023-10-26
PI-1585 Added update message processing

stevomcallister:


projects/cas3-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/ContactService.kt
```diff
@@ -30,15 +32,30 @@ class ContactService(
...
+        val existing = contactRepository.getByExternalReference(event.eventDetails.urn)
+        if (existing != null) {
+            if (existing.startTime < event.timestamp) {
+                existing.notes += "\n${event.eventDetails.noteText}"
+                existing.date = event.timestamp.toLocalDate()
+                existing.startTime = event.timestamp
+                contactRepository.save(existing)
+            } else {
+                telemetryService.trackEvent(
+                    "Ignoring out of sequence older message.",
```
marcus-bcl, line 44: I guess this means we potentially lose information e.g. if contact B arrives before contact A, then we ignore the content of A and can only see the amendment from B.

Handling that might be more effort than it's worth though - I assume they have a link back to the AP service somewhere to view the full history.

marcus-bcl (approved):

projects/cas3-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/ContactService.kt
marcus-bcl: (What he said https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/2452#discussion_r1373117576 😄)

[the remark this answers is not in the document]
stevomcallister: After discussions with the teams - have decided to only store the latest from CAS3 and not to try to store a history

stevomcallister: After discussions with the teams - have decided to only store the latest from CAS3 and not to try to store a history

PR #2475, opened 2023-10-31
PI-1602 additional fix for duplicate addresses

stevomcallister:


marcus-bcl (commented):
Could this still potentially create duplicates if they don't already have a main address to start with?

marcus-bcl (approved):

PR #2487, opened 2023-11-02
PI-1614

anthony-britton-moj:


projects/manage-pom-cases-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/services/PrisonManagerService.kt
```diff
@@ -39,12 +39,31 @@ class PrisonManagerService(
         val staff = getStaff(probationArea, team, allocation.manager, allocationDate)
         val currentPom = prisonManagerRepository.findActiveManagerAtDate(personId, allocationDate)
         val newPom = currentPom.changeTo(personId, allocationDate, probationArea, team, staff)
+        val newEndDate = if (currentPom == null) {
+            prisonManagerRepository.findFirstManagerAfterDate(personId, allocationDate).firstOrNull()?.date
+        } else {
+            null
+        }
```
marcus-bcl, line 46: Are we still missing the normal historical allocation here?  If the currentPom has an endDate, we should use that as newEndDate, and then we should update the currentPom.endDate to be allocationDate ... I think 🤔 

```suggestion
        val newEndDate = currentPom?.endDate ?: prisonManagerRepository.findFirstManagerAfterDate(personId, allocationDate).firstOrNull()?.date
        currentPom?.endDate = allocationDate
```

stevomcallister (dismissed):

PR #2490, opened 2023-11-03
Document Probation System Concepts 

mxco86:
- Initial attempt at starting an overview of the concepts we use in the integration services
- Also classify the integration services into service areas

### DIFF
--- a/doc/tech-docs/source/concepts.html.md.erb
+++ b/doc/tech-docs/source/concepts.html.md.erb
@@ -1,0 +1,71 @@
+---
+title: Probation System Concepts
+source_url: 'https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/doc/tech-docs/source/concepts.html.md.erb'
+weight: 40
+---
+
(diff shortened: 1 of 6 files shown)


doc/tech-docs/source/concepts.html.md.erb
```diff
@@ -0,0 +1,71 @@
+---
+title: Probation System Concepts
+source_url: 'https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/doc/tech-docs/source/concepts.html.md.erb'
+weight: 50
+---
+
+# System Concepts
```
stevomcallister, line 7: What about Nomis/prison-api? Alfresco?

mxco86: Was scoping this to just the things the we provide so that would exclude NOMIS concepts. Seems worth adding some Alfresco information to the Documents section

marcus-bcl (approved):

PR #2503, opened 2023-11-06
PI-1629 check for non-establishments

anthony-britton-moj:


projects/soc-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationAreaEntity.kt
```diff
@@ -95,6 +97,7 @@ interface ProbationAreaRepository : JpaRepository<ProbationAreaEntity, Long> {
         and pa.selectable = true
         and d.selectable = true
         and b.selectable = true
+        and pa.establishment is null
```
marcus-bcl, line 100: Do we need to add the same check to the `probationAreaDistrictsNonSelectable` method?  Also would checking for not equal to Y be better?  I know they're all either Y or null in the DB at the moment, but just in case someone decides to create one with establishment=N.

marcus-bcl (dismissed):

stevomcallister (approved):

marcus-bcl (approved):

PR #2725, opened 2023-11-23
PI-1619 make endpoint return all crns in one go

anthony-britton-moj:


projects/tier-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/PersonRepository.kt
```diff
@@ -2,11 +2,10 @@ package uk.gov.justice.digital.hmpps.integrations.delius.person
 
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
-import java.util.stream.Stream
 
 interface PersonRepository : JpaRepository<Person, Long> {
     fun findByCrnAndSoftDeletedIsFalse(crn: String): Person?
 
     @Query("select p.crn from Person p where p.softDeleted = false")
```
marcus-bcl, line 9: Not related to this PR, but do we need to worry about CRNs without an active event?

marcus-bcl (approved):

projects/tier-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/PersonRepository.kt
marcus-bcl: No worries if that's what we decided, was just wondering - the full dataset isn't that big anyway.

stevomcallister (approved):

PR #2870, opened 2023-12-05
Make username lookup case-insensitive

marcus-bcl:


marcus-bcl:
> feels like we've done this a lot ... is there many more places we do a user lookup that don't already have he case insensitive version?

From what I can tell this is the only one.  It's the one in the audit lib that's been there forever, so I don't think it's actually causing any issues right now.

PR #2905, opened 2023-12-12
PI-1643 Endpoint for updating custody key dates via noms ids

stevomcallister:


projects/custody-key-dates-and-delius/src/main/resources/application.yml
```diff
@@ -44,6 +44,7 @@ spring.config.activate.on-profile: [ "dev", "integration-test" ]
 spring:
   datasource.url: jdbc:h2:file:./dev;MODE=Oracle;DEFAULT_NULL_ORDERING=HIGH;AUTO_SERVER=true;AUTO_SERVER_PORT=9092
   jpa.hibernate.ddl-auto: create-drop
+  security.oauth2.resourceserver.jwt.public-key-location: classpath:local-public-key.pub
```
marcus-bcl, line 47: You'll need to add the extra auth env vars to the values files too to make this work as an API server e.g. https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/projects/approved-premises-and-delius/deploy/values-dev.yml#L11-L12

marcus-bcl (dismissed):
Looks good 👍

PR #2910, opened 2023-12-12
PI-1643 fixed issue with lowercase noms numbers

stevomcallister:


marcus-bcl (approved):
😨 Do we have lowercase NOMIS numbers in Delius?

stevomcallister:
> 😨 Do we have lowercase NOMIS numbers in Delius?

we do in preprod

PR #2949, opened 2023-12-20
PI-1752 update workforce to record telemetry rather than exceptions

anthony-britton-moj:
for common errors


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/allocations/AllocationValidator.kt
```diff
@@ -27,7 +27,7 @@ class AllocationValidator(
             ?: throw NotFoundException("Team", "code", allocationDetail.teamCode)
 
         if (team.providerId != providerId) {
-            throw ConflictException("Cannot transfer from provider $providerId to ${team.providerId}")
+            throw IgnorableMessageException("External transfer not permitted")
```
marcus-bcl, line 30: Should we include the from/to provider in the telemetry properties?

marcus-bcl (approved):

PR #2973, opened 2024-01-03
Pi 1744 office address search

pmcphee77:


projects/create-and-vary-a-licence-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/OfficeAddressIntegrationTest.kt
```diff
@@ -0,0 +1,85 @@
...
+import uk.gov.justice.digital.hmpps.test.MockMvcExtensions.withToken
+
+@AutoConfigureMockMvc
+@SpringBootTest(webEnvironment = RANDOM_PORT)
+internal class OfficeAddressIntegrationTest {
+    @Autowired
+    lateinit var mockMvc: MockMvc
+
+    @Autowired
+    lateinit var objectMapper: ObjectMapper
```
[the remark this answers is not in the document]
pmcphee77: Thanks! Thats much better - updated

pmcphee77:
> Looks like the coverage is under the required amount still? Should really be 80% or above - current sonar shows only 66.7% on new code. Might be worth excluding the Office Address model to improve this



> Looks like the coverage is under the required amount still? Should really be 80% or above - current sonar shows only 66.7% on new code. Might be worth excluding the Office Address model to improve this

Ive removed the Companion object and from method as i dont belive i want to be returning any nulls actually. I took that logic from the address endpoint but that wasnt a list

projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/provider/entity/Provider.kt
```diff
@@ -108,3 +110,47 @@ interface BoroughRepository : JpaRepository<Borough?, Long?> {
...
+    @JoinColumn(name = "district_id")
+    @ManyToOne
+    val district: District,
+
+    @Id
+    @Column(name = "office_location_id")
+    val id: Long
+)
+
+interface OfficeLocationRepository : JpaRepository<OfficeLocation?, Long?> {
```
marcus-bcl, line 146: Should these types be nullable here?  Not sure whether it actually makes a difference

pmcphee77: yeah no reason for these being null. Not sure why i did that. Ive changed. Thanks

pmcphee77: ive also changed the other instance in that class

marcus-bcl (approved):

PR #2985, opened 2024-01-10
PI-1533: Added staff caseload endpoint

pmcphee77:


projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/ManagedOffender.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.api.model
+
+import java.time.LocalDate
+
+data class ManagedOffender(
+    val offenderCrn: String,
+    val offenderName: Name,
```
marcus-bcl, line 7: We probably don't need the `offender` prefix here
```suggestion
    val crn: String,
    val name: Name,
```

projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/resource/StaffResource.kt
```diff
@@ -31,4 +32,9 @@ class StaffResource(
         staffService.findStaffForUsernames(
             usernames
         )
+
+    @PreAuthorize("hasRole('CVL_CONTEXT')")
+    @GetMapping("/{staffCode}/caseload/managed-offenders")
+    fun getManagedOffenders(@PathVariable staffCode: String): List<ManagedOffender> =
```
marcus-bcl, line 38: The Community API version was paginated - do we need to do the same here?  Sorry if it's already been discussed.

pmcphee77: The current community API is not currently paginated. See https://community-api-public.test.probation.service.justice.gov.uk/swagger-ui/index.html#/Staff/getCaseloadOffendersForStaff - i discussed this with Anthony and we decided not to introduce this.

projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/ManagedOffender.kt
pmcphee77: Done

projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/resource/StaffResource.kt
marcus-bcl: My mistake - I was looking at the team caseload endpoint.

marcus-bcl (approved):

PR #3008, opened 2024-01-11
PI-1785: Added get user details by id

pmcphee77:
Ive had to add a /user/details/{userId} endpoint as user/{username} already in use. Also can't easily add query param as /user/?email already in use which (apart from being messy to introduce a userId query param) currently returns a list. We want a 404 or single user object

### DIFF
--- a/projects/hmpps-auth-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
+++ b/projects/hmpps-auth-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
@@ -7,6 +7,7 @@ import org.springframework.ldap.query.LdapQueryBuilder.query
 import org.springframework.ldap.query.SearchScope
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.entity.LdapUser
+import uk.gov.justice.digital.hmpps.entity.UserRepository
 import uk.gov.justice.digital.hmpps.exception.NotFoundException
 import uk.gov.justice.digital.hmpps.ldap.byUsername
 import uk.gov.justice.digital.hmpps.ldap.findByUsername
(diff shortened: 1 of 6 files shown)


projects/hmpps-auth-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -60,4 +67,14 @@ class UserService(
         enabled = enabled,
         roles = getUserRoles(dn)
     )
+
+    private fun LdapUser.toUserDetails(userId: Long) = UserDetails(
```
marcus-bcl, line 71: Is it worth combining this with the method above to reduce repetition?  Or even calling this one from the other e.g. something like:
```kotlin
    private fun LdapUser.toUserDetails() = auditUserService.findUser(username)?.let { toUserDetails(it.id) } 
          ?: throw NotFoundException("User entity", "username", username)
```

projects/hmpps-auth-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -17,11 +18,17 @@ import javax.naming.Name
 @Service
 class UserService(
     private val ldapTemplate: LdapTemplate,
-    private val auditUserService: AuditUserService
+    private val auditUserService: AuditUserService,
```
marcus-bcl, line 21: Do you think we should move the existing calls to AuditUserService over to the new UserRepository for consistency?

projects/hmpps-auth-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
pmcphee77: Yeah - absolutely. Should have done this. Thanks. Done

projects/hmpps-auth-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
pmcphee77: Yeah - much better - ive moved them and not using AuditRepository anymore. Done

marcus-bcl (approved):

PR #3054, opened 2024-01-19
PI-1811

anthony-britton-moj:


projects/refer-and-monitor-and-delius/src/test/kotlin/uk/gov/justice/digital/hmpps/service/NsiServiceTest.kt
```diff
@@ -23,7 +23,9 @@ import uk.gov.justice.digital.hmpps.integrations.delius.referral.NsiStatusHistor
 import uk.gov.justice.digital.hmpps.integrations.delius.referral.NsiStatusRepository
 import uk.gov.justice.digital.hmpps.integrations.delius.referral.entity.NsiStatus
 import uk.gov.justice.digital.hmpps.telemetry.TelemetryService
+import java.time.LocalDate
```
marcus-bcl, line 26: Are these imports used?

marcus-bcl (dismissed):

projects/refer-and-monitor-and-delius/src/test/kotlin/uk/gov/justice/digital/hmpps/service/NsiServiceTest.kt
marcus-bcl: Shame we can't auto-fix these in the CLI: https://youtrack.jetbrains.com/issue/IDEA-242015/Optimize-Imports-via-CLI

marcus-bcl (approved):

PR #3116, opened 2024-01-24
PI-1828

anthony-britton-moj:


projects/approved-premises-and-delius/src/main/resources/application.yml
```diff
@@ -46,7 +46,9 @@ spring.config.activate.on-profile: [ "dev", "integration-test" ]
 
 spring:
   datasource.url: jdbc:h2:file:./dev;MODE=Oracle;DEFAULT_NULL_ORDERING=HIGH;AUTO_SERVER=true;AUTO_SERVER_PORT=9092
-  jpa.hibernate.ddl-auto: create-drop
+  jpa:
+    hibernate.ddl-auto: create-drop
+    database-platform: org.hibernate.dialect.OracleDialect
```
marcus-bcl, line 51: Oops, sorry - I thought I'd added this back in everywhere but looks like I missed it for this one service. Probably worth moving to the top so it applies to all profiles. Here's where I accidentally removed it: https://github.com/ministryofjustice/hmpps-probation-integration-services/pull/3035/files#diff-bba0eb0cb40c0379c76fd5ee9dfabf4f76c73fa5b60a9fb3225aaaab9c8183d3L7

.github/workflows/reports.yml
```diff
@@ -54,16 +55,108 @@ jobs:
...
+          echo "result=$(curl -fsSL -H "x-api-key: $key" --data-urlencode "query=$query" --get "$url")" | tee -a "$GITHUB_OUTPUT"
+        env:
+          url: https://api.applicationinsights.io/v1/apps/${{ secrets.APP_INSIGHTS_APPLICATION_GUID }}/query
+          key: ${{ secrets.APP_INSIGHTS_API_KEY }}
+          query: |
+            customEvents
+            | where timestamp > ago(7d) and timestamp < now()
+            | where cloud_RoleName in ("approved-premises-and-delius")
+            | where name in ("ApprovedPremisesFailureReport")
+            | project customDimensions.crn, customDimensions.reason, customDimensions.externalReference, itemId, timestamp
```
marcus-bcl, line 112: Do you need the externalReference here?  Or are you planning to include it in the table?

marcus-bcl (approved):

PR #3129, opened 2024-01-29
PI-1834: Add team addresses to responsible com endpoint

pmcphee77:
Ive only put this in the team object when the team is as part of the manager endpoint


projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/manager/entity/PersonManager.kt
```diff
@@ -34,6 +30,15 @@ class PersonManager(
...
     val team: Team,
 
+    @ManyToMany
+    @JoinTable(
+        name = "team_office_location",
+        joinColumns = [JoinColumn(name = "team_id")],
+        inverseJoinColumns = [JoinColumn(name = "office_location_id")]
+    )
+    @SQLRestriction("end_date is null or end_date > current_date")
+    val teamOfficeLocations: List<OfficeLocation>?,
```
marcus-bcl, line 40: Does this need to be nullable, or would it come back as an empty list if there were no office locations?

marcus-bcl (approved):

marcus-bcl (approved):

PR #3250, opened 2024-02-14
PI-1907

anthony-britton-moj:


projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/event/entity/Disposal.kt
```diff
@@ -43,12 +43,7 @@ class Disposal(
 
     @Column(updatable = false, columnDefinition = "NUMBER")
     val softDeleted: Boolean = false
-) {
-    fun isLongerThan20Months(): Boolean? {
```
marcus-bcl, line 47: Not related to the PR - but just wondered why we stopped using this check?   Aren't we only supposed to create a POM if they're on a sentence of 20 months or longer, or something like that?

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #3261, opened 2024-02-20
PI-1929 Explicitly disable audit user creation in preprod

marcus-bcl:
This allows us to continue creating the client_id trigger.

See https://github.com/ministryofjustice/hmpps-delius-pipelines/pull/793


bill-buchan (approved):
LGTM

PR #3314, opened 2024-02-21
PI-1944

anthony-britton-moj:


projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ReferralService.kt
```diff
@@ -164,15 +164,23 @@ class ReferralService(
...
+    fun getReferral(person: Person, externalReference: String): Referral {
+        val referral: Referral = referralRepository.findByPersonIdAndCreatedByUserIdAndReferralNotesContains(
             person.id,
             ServiceContext.servicePrincipal()!!.userId,
             externalReference
         ) ?: throw IgnorableMessageException(
             "Referral Not Found",
             mapOf("crn" to person.crn, "externalReference" to externalReference)
         )
+        if (referral.approvedPremisesId == null) {
```
pmcphee77, line 176: Is there a test for this? Or do we not need?

pmcphee77 (approved):

PR #3318, opened 2024-02-21
PI-1945 Fix handling of large descriptions with new lines in Trivy issues

marcus-bcl:


.github/workflows/security.yml
```diff
@@ -80,38 +77,41 @@ jobs:
...
           | while read -r vulnerability; do
-            title=$(echo "$vulnerability" | jq -r '.Title')
-            body=$(echo "$vulnerability" | jq -r '.Body')
+            title=$(jq -r '.Title' <<< "$vulnerability")
+            jq -r '.Body' <<< "$vulnerability" | tee "${title}-body.json"
             existing=$(gh issue list --state open --label dependencies --label security --search "$title" --json 'number' --jq '.[].number' | head -n1)
             if [ -n "$existing" ]; then
               echo "Issue '$title' already exists, updating body..."
-              gh issue edit "$existing" --body "$body"
+              echo gh issue edit "$existing" --body-file "${title}-body.json"
```
mxco86, line 103: Is the `echo` supposed to be in here?

.github/workflows/security.yml
```diff
@@ -80,38 +77,41 @@ jobs:
...
             else
-              gh issue create --title "$title" --body "$body" --label 'dependencies,security'
```
mxco86, line 104: And here?

.github/workflows/security.yml
marcus-bcl: Oops - good spot, accidentally added that back in when fixing the linting issues.

mxco86 (approved):

PR #3400, opened 2024-02-29
PI-1965 Add Extra Status Detail to CAS2 Contacts

achimber-moj:


marcus-bcl (commented):
Looks good, but are we missing the part to add the new detail into the notes of the entity we're creating?

We construct the notes string based on the event details here: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/e0502bf04eee9e1cc5a9be14b033e070af8c2c40/projects/cas2-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/Cas2Service.kt#L41


Also - I hope you don't mind but I added the ticket number to the PR title, so that it all links up in Jira 🙂

marcus-bcl (approved):

marcus-bcl (approved):

PR #3427, opened 2024-03-05
PI-1961

anthony-britton-moj:
WIP - awaiting further information to complete

### DIFF
--- a/projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Appointment.kt
+++ b/projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Appointment.kt
@@ -59,8 +59,10 @@ class Appointment(
     @JoinColumn(name = "staff_id")
     val staff: Staff,
 
+    val externalReference: String? = null,
+
     @Column(name = "description")
-    val description: String?,
+    val description: String? = null,
 
     @ManyToOne
     @JoinColumn(name = "office_location_id")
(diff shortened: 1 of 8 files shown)


projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Appointment.kt
```diff
@@ -144,4 +179,43 @@ interface AppointmentRepository : JpaRepository<Appointment, Long> {
...
+
+    @Query(
+        """
+            select count(c.contact_id)
+            from contact c
+            join r_contact_type ct on c.contact_type_id = ct.contact_type_id
+            where c.offender_id = :personId and ct.attendance_contact = 'Y'
+            and to_char(c.contact_date, 'YYYY-MM-DD') = :date
+            and to_char(c.contact_start_time, 'HH24:MI') < :endTime 
+            and to_char(c.contact_end_time, 'HH24:MI') > :startTime
```
marcus-bcl, line 191: Should this be an OR check, to account for any overlap whether it's at the start or end?
```suggestion
            and (to_char(c.contact_start_time, 'HH24:MI') < :endTime or to_char(c.contact_end_time, 'HH24:MI') > :startTime)
```

marcus-bcl (approved):

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Appointment.kt
marcus-bcl: Sorry - ignore me!

PR #3459, opened 2024-03-12
Pi 1991 personal details

pmcphee77:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonalDetailsService.kt
```diff
@@ -0,0 +1,106 @@
...
+            mobileNumber = person.mobileNumber,
+            circumstances = Circumstances(lastUpdated = person.personalCircumstances.maxByOrNull { it.lastUpdated }?.lastUpdated,
+                circumstances = person.personalCircumstances.map {
+                    PersonalCircumstance(
+                        it.subType.description,
+                        it.type.description
+                    )
+                }),
+            disabilities = Disabilities(
+                lastUpdated = person.disabilities.maxByOrNull { it.lastUpdated }?.lastUpdated,
```
marcus-bcl, line 51: Minor thing - you should be able to replace thes `maxByOrNull`s with `maxOfOrNull`
```suggestion
                lastUpdated = person.disabilities.maxOfOrNull { it.lastUpdated },
```

marcus-bcl (approved):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonalDetailsService.kt
pmcphee77: Done

marcus-bcl (approved):

PR #3465, opened 2024-03-14
PI-1957

anthony-britton-moj:
Use async to improve performance
Limit concurrency to control memory usage


marcus-bcl (approved):
Why do we need the limit out of interest?  Does AWS not handle too many concurrent requests?

PR #3596, opened 2024-04-08
Feature/pi 1939 mas probabtion history

achimber-moj:


projects/manage-supervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/SentenceIntegrationTest.kt
```diff
@@ -67,8 +67,8 @@ class SentenceIntegrationTest {
                     Order("Default Sentence Type", 12, null, LocalDate.now().minusDays(14)),
                     listOf(Requirement("Main", "High Intensity", 12, "my notes", Rar(1, 0, 1))),
                     listOf(
-                        CourtDocument(115, LocalDate.now().minusDays(1), "court report"),
-                        CourtDocument(116, LocalDate.now().minusDays(3), "event report")
+                        CourtDocument(118, LocalDate.now().minusDays(1), "court report"),
```
marcus-bcl, line 70: It looks like these CourtDocument entities are generated here: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/projects/manage-supervision-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/DataLoader.kt#L75

It might be cleaner if we moved those to some static vals (e.g. in a CourtDocumentGenerator), so we could reference them here, rather than updating these ids each time?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/personalDetails/entity/PersonalContactEntity.kt
```diff
@@ -76,9 +79,35 @@ interface PersonalContactRepository : JpaRepository<PersonalContactEntity, Long>
...
 
 fun PersonalContactRepository.getContact(crn: String, id: Long): PersonalContactEntity =
     findById(crn, id) ?: throw NotFoundException("PersonalContact", "id", id)
 
+interface ContactDetails {
+    val firstName: String
+    val otherNames: String?
+    val surname: String
+    val mobileNumber: String
+    val emailAddress: String
```
marcus-bcl, line 110: Aren't these fields nullable?

pmcphee77 (dismissed):

marcus-bcl (approved):

pmcphee77 (approved):

PR #3598, opened 2024-04-08
PI-2054: Added compliance and activity

pmcphee77:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/compliance/Nsi.kt
```diff
@@ -0,0 +1,74 @@
...
+    @ManyToOne
+    @JoinColumn(name = "nsi_status_id")
+    val nsiStatus: NsiStatus? = null,
+
+    @Id
+    @Column(name = "nsi_id")
+    val id: Long = 0,
+
+    @Column(name = "active_flag", columnDefinition = "number")
+    var active: Boolean = true,
```
marcus-bcl, line 40: Is this intended to be mutable?
```suggestion
    val active: Boolean = true,
```

marcus-bcl (approved):

marcus-bcl (approved):

PR #3623, opened 2024-04-12
PI-2019 Update production ingress class and replica count

marcus-bcl:
To resolve cross-zone load balancing issues


pmcphee77 (dismissed):

marcus-bcl:
Note: waiting on confirmation from Cloud Platform team that the production-only ingress is ready for use.

marcus-bcl:
Update - This is on the MOJ Cloud Platform roadmap for Q3 2024.

PR #3629, opened 2024-04-15
PI-1939 - handle null code_description from r_standard_reference_list…

achimber-moj:
… table


achimber-moj:
> Code Description should not be null in reference data Probably worth a catch up to see what is actually happening here

for curfew cases the code description is not set.  i have a left join for these cases. 

crn                    code          description                                          code description 
D002366	7	           Court - Accredited Programme	     ASRO	
D002366	RM49	  Curfew                                                (Electronic Monitored)		
D002366	Y	          Supervision	                                     Regular Supervision	
D002366	W	          Unpaid Work	                                     Regular

happy to have a get together.

PR #3660, opened 2024-04-22
PI-2065: Return caseload information

pmcphee77:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
```diff
@@ -14,9 +14,146 @@ class User(
...
+    @Column(name = "probation_area_id")
+    val id: Long,
+
+    @Column(name = "end_date")
+    val endDate: LocalDate? = null
+)
+
+@Entity
+@Immutable
+@Table(name = "case_load")
```
marcus-bcl, line 111: Is this correct?  I think the table name is `caseload`.

Also, I wonder if we should be using the caseload table at all?  I know we've had issues in the past where it hasn't been kept up-to-date with the source tables (`offender_manager`, `order_manager` etc), but it is what the Delius caseload diary uses...

pmcphee77: ive corrected the table name to caseload. Ive also added an SQL restriction for now to only return role_code = 'OM'

marcus-bcl (approved):

PR #3669, opened 2024-04-23
PI-2065: Updated caseload api

pmcphee77:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
```diff
@@ -29,6 +30,23 @@ class User(
...
 
+interface UserRepository : JpaRepository<User, Long> {
+
+    @Query(
+        """
+        select u
+        from User u
+        join fetch u.staff s
+        join fetch s.provider p
+        where u.username = :username
```
marcus-bcl, line 41: Username is case-insensitive in Delius, and there's an index on `upper(username)`.
```suggestion
        where upper(u.username) = upper(:username)
```

marcus-bcl (approved):

PR #3678, opened 2024-04-25
Pi 2124 find ap documents

pmcphee77:


projects/make-recall-decisions-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/entity/Entities.kt
```diff
@@ -0,0 +1,24 @@
+package uk.gov.justice.digital.hmpps.data.entity
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.time.LocalDate
+
+@Entity
+class ApprovedPremisesReferral(
```
marcus-bcl, line 10: Guess we don't need this any more?

marcus-bcl (approved):

projects/make-recall-decisions-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/entity/Entities.kt
pmcphee77: Ive removed it

marcus-bcl (approved):

PR #3732, opened 2024-05-10
Feature/pi 2158 mas get unpaid work time

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceService.kt
```diff
@@ -116,6 +119,22 @@ class SentenceService(
         return null
     }
 
+    fun getUnpaidWorkTime(requirementType: String, eventId: Long, eventNumber: String): String? {
+        if (requirementType.equals("W", true)) {
+            val durationInMinutes: Long = upwAppointmentRepository.calculateUnpaidTimeWorked(eventId, eventNumber)
+            durationInMinutes.let {
+                val durationInHours = it / 60
```
pmcphee77, line 126: do we need to think about rounding here?

pmcphee77 (changes requested):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceService.kt
achimber-moj: will change the logic for this one

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/sentence/Requirement.kt
```diff
@@ -13,6 +13,7 @@ data class Requirement(
     val description: String,
     val length: Long?,
     val lengthUnitValue: String?,
+    val progress: String?,
```
marcus-bcl, line 16: Could we call this something more UPW-specific?  E.g. `unpaidWorkProgress`?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/sentence/entity/UpwAppointment.kt
```diff
@@ -0,0 +1,45 @@
...
+                SELECT ua.MINUTES_CREDITED AS duration
+                FROM UPW_APPOINTMENT ua
+                JOIN OFFENDER o 
+                ON o.OFFENDER_ID = ua.OFFENDER_ID 
+                JOIN EVENT e 
+                ON e.OFFENDER_ID = o.OFFENDER_ID 
+                WHERE e.EVENT_ID = :id
+                AND e.EVENT_NUMBER = :eventNumber
+                AND ua.ATTENDED = 'Y'
+                AND ua.SOFT_DELETED = 0)
```
marcus-bcl, line 41: I'm not sure you need this sub-select.  Also, this looks like it'll get any UPW appointments for the offender, even if the appointments are for other events?  You might be able to link to the sentence (disposal) via the upw_details, something like:

```suggestion
            select sum(minutes_credited) 
            from upw_appointment 
            join upw_details on upw_details.upw_details_id = upw_appointment.upw_details_id
            join disposal on disposal.disposal_id = upw_details.disposal_id
            where disposal.event_id = :eventId
            and upw_appointment.attended = 'Y'
            and upw_appointment.soft_deleted = 0
```

However at that point, are we sure we should be returning this as part of the Requirement model - if it's actually linked to the Sentence?

achimber-moj: sql updated.  i am happy to leave the model at the moment

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceService.kt
achimber-moj: updated logic

PR #3735, opened 2024-05-10
PI-2189 Move ou=Users into LDAP base config

marcus-bcl:


pmcphee77 (commented):
Looks like the coverage check is failing

pmcphee77 (approved):

PR #3740, opened 2024-05-13
PI-1915: Added probation-case/crn endpoint

pmcphee77:


projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/Disability.kt
```diff
@@ -0,0 +1,56 @@
...
+    val condition: ReferenceData,
+
+    val startDate: LocalDate,
+
+    @Column(name = "last_updated_datetime")
+    val lastUpdated: ZonedDateTime,
+
+//    @ManyToOne
+//    @JoinColumn(name = "last_updated_user_id")
+//    val lastUpdatedUser: User,
```
marcus-bcl, line 42: Can this be removed?

marcus-bcl (approved):

PR #3804, opened 2024-05-16
PI-2194: Added new event for recall consideration

pmcphee77:


projects/make-recall-decisions-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/RecommendationService.kt
```diff
@@ -36,20 +36,45 @@ class RecommendationService(
...
+    ) {
+        val contact = addContact(crn, null, details, username, occurredAt, ContactType.CONSIDERATION)
+        contactRepository.save(contact)
+        telemetryService.trackEvent(
+            "ConsiderationCreated",
+            mapOf("CRN" to crn, "username" to username)
+        )
+    }
+
+    private fun addContact(
```
marcus-bcl, line 61: Minor thing - but could combine this with the Person.addContact function at the bottom of the class?

pmcphee77: Think they both do different things. The original one doesnt actual add anyting (it gets a contact using person) so ive changed that to getPerson to make it clearer

achimber-moj (approved):

PR #3811, opened 2024-05-20
PI-2209 filter ocg alerts

achimber-moj:


projects/prison-case-notes-to-probation/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/service/DeliusService.kt
```diff
@@ -24,21 +27,50 @@ class DeliusService(
...
+    private val ignoreMessage: String?,
 ) : AuditableService(auditedInteractionService) {
     @Transactional
     fun mergeCaseNote(@Valid caseNote: DeliusCaseNote) = audit(CASE_NOTES_MERGE) {
         it["nomisId"] = caseNote.header.noteId
 
         val existing = caseNoteRepository.findByNomisId(caseNote.header.noteId)
 
-        val entity = if (existing == null) caseNote.newEntity() else existing.updateFrom(caseNote)
+        val entity = if (existing == null) checkOCGMessages(caseNote) else existing.updateFrom(caseNote)
```
marcus-bcl, line 41: Will this only apply to new case notes?  Do we also want to ignore any updates that contain OCG information?

Also this would be better placed in the PrisonCaseNoteFilters class, there are some existing predicates in there that we use to filter out case notes.  That would also handle the logging to Telemetry for you.

projects/prison-case-notes-to-probation/src/main/resources/application.yml
```diff
@@ -39,6 +39,8 @@ management:
 logging.level:
   org.hibernate.engine.jdbc.spi.SqlExceptionHelper: OFF
 
+ignore.message: NOMIS Alert Security. Do not share with offender and OCG Nominal - Do not share made active.
```
marcus-bcl, line 42: We're probably fine to hard-code this in the Kotlin code for simplicity - it seems like it will be relatively static in each environment.

projects/prison-case-notes-to-probation/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/service/DeliusService.kt
achimber-moj: This will only apply to new case notes.  I did consider adding at the handler layer, but happy where it is at, unless you would prefer it t be moved.

projects/prison-case-notes-to-probation/src/main/resources/application.yml
achimber-moj: will update

projects/prison-case-notes-to-probation/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/service/DeliusService.kt
achimber-moj: done

projects/prison-case-notes-to-probation/src/main/resources/application.yml
achimber-moj: done

projects/prison-case-notes-to-probation/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/prison/PrisonCaseNoteFilters.kt
```diff
@@ -7,7 +7,10 @@ object PrisonCaseNoteFilters {
         },
         PrisonCaseNoteFilter("Prisoner being transferred") {
             it.locationId == "TRN"
-        }
+        },
+        PrisonCaseNoteFilter("OCG Alert Security") {
+            it.text == "Alert Security. Do not share with offender and OCG Nominal - Do not share made active."
```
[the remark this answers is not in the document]
achimber-moj: updated

projects/prison-case-notes-to-probation/src/test/kotlin/uk/gov/justice/digital/hmpps/messaging/HandlerTest.kt
```diff
@@ -99,8 +101,12 @@ internal class HandlerTest {
         verify(telemetryService).trackEvent(eq("CaseNoteIgnored"), any(), any())
     }
 
-    @Test
-    fun `when ocg alert text has been sent`() {
+    @ParameterizedTest
+    @ValueSource(
```
marcus-bcl, line 105: Nice! I didn't know about the `@ValueSource` annotation, looks like it makes parameterized tests a bit simpler to write.

marcus-bcl (approved):

PR #3828, opened 2024-05-22
PI-2161: Created post endpoint for /nomis-case-note/{crn}

pmcphee77:


projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/CreateContact.kt
```diff
@@ -0,0 +1,22 @@
...
+    )
+    val type: Type,
+    val dateTime: ZonedDateTime,
+    @field:NotBlank
+    val notes: String,
+) {
+    enum class Type {
+        RP9,
+        RP10,
+    }
```
marcus-bcl, line 21: Sorry just thought of this - it seems like RP9 & RP10 are internal Delius codes, maybe we should map them from a friendlier name instead so that Resettlement Passport doesn't need to map it themselves?

For example,
```suggestion
    enum class CaseNoteType(contactTypeCode: String) {
        IMMEDIATE_NEEDS_REPORT("RP9"),
        PRE_RELEASE_REPORT("RP10"),
    }
```

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/CaseNote.kt
```diff
@@ -0,0 +1,98 @@
...
+    val type: CaseNoteType,
+
+    @Lob
+    var notes: String,
+
+    @Column(name = "contact_date")
+    var date: LocalDate,
+
+    @Column(name = "contact_start_time")
+    var startTime: ZonedDateTime,
```
marcus-bcl, line 39: Should these be `val`s?  There's a few more further down in the entity too.

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseNoteService.kt
```diff
@@ -0,0 +1,35 @@
...
+@Service
+class CaseNoteService(
+    auditedInteractionService: AuditedInteractionService,
+    private val caseNoteRepository: CaseNoteRepository,
+    private val personRepository: PersonRepository,
+    private val personManagerRepository: PersonManagerRepository,
+    private val caseNoteTypeRepository: CaseNoteTypeRepository
+) : AuditableService(auditedInteractionService) {
+
+    fun createContact(crn: String, createContact: CreateContact) {
```
marcus-bcl, line 18: This is extending the AuditableService, but not doing any auditing.  You just need to wrap your method call with a call to `audit` e.g.
```suggestion
) : AuditableService(auditedInteractionService) {

    fun createContact(crn: String, createContact: CreateContact) = audit(BusinessInteractionCode.ADD_CONTACT) {
```

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/CreateContact.kt
pmcphee77: yeah i did think of this actually. Looks like more changes may be required anyway

pmcphee77: Made these changes along with others

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/CaseNote.kt
pmcphee77: fixed

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseNoteService.kt
pmcphee77: done!

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Staff.kt
```diff
@@ -20,14 +17,18 @@ class Staff(
...
-    val user: StaffUser?,
+    val user: StaffUser? = null,
+
+    val probationAreaId: Long,
 
     @Id
     @Column(name = "staff_id")
-    val id: Long
+    @SequenceGenerator(name = "staff_id_seq", sequenceName = "staff_id_seq", allocationSize = 1)
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "staff_id_seq")
```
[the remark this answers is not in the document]
pmcphee77: done and moved to new library

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/StaffTeam.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.data.annotation.LastModifiedDate
+import org.springframework.data.jpa.domain.support.AuditingEntityListener
+import java.io.Serializable
+import java.time.ZonedDateTime
+
+@EntityListeners(AuditingEntityListener::class)
+@Entity
+@Table(name = "staff_team")
+@IdClass(StaffTeamId::class)
+class StaffTeam(
```
marcus-bcl, line 16: Do you still need to define the Staff and StaffTeam entities in this project?

If they're still used for the appointment endpoint, I guess we can make them `@Immutable` here and remove the created/lastUpdated columns now?

marcus-bcl (approved):
Looks good!

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/StaffTeam.kt
pmcphee77: yeah i did need to define them as i wanted to assert they have been created properly with all the fields in the tests

PR #3837, opened 2024-05-29
PI-1803 Add components to HMPPS Service Catalogue on bootstrap

marcus-bcl:
Depends on:
* https://github.com/ministryofjustice/hmpps-github-discovery/pull/36
* https://github.com/ministryofjustice/hmpps-service-catalogue/pull/85

I've ran this for all existing projects here:
* Logs: https://github.com/ministryofjustice/hmpps-probation-integration-services/actions/runs/9304322564
* Result: https://developer-portal.hmpps.service.justice.gov.uk/products/185

### DIFF
--- a/projects/court-case-and-delius/src/main/resources/application.yml
+++ b/projects/court-case-and-delius/src/main/resources/application.yml
@@ -74,9 +74,7 @@ spring.datasource.url: jdbc:h2:mem:./test;MODE=Oracle;DEFAULT_NULL_ORDERING=HIGH
 
 ---
 spring.config.activate.on-profile: oracle
-spring.datasource:
-  url: 'jdbc:tc:oracle:slim-faststart:///XEPDB1'
-  jpa.hibernate.ddl-auto: create
+spring.datasource.url: 'jdbc:tc:oracle:slim-faststart:///XEPDB1'
 
 ---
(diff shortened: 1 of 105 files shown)


achimber-moj (approved):

projects/court-case-and-delius/src/main/resources/application.yml
```diff
@@ -72,7 +74,9 @@ spring.datasource.url: jdbc:h2:mem:./test;MODE=Oracle;DEFAULT_NULL_ORDERING=HIGH
 
 ---
 spring.config.activate.on-profile: oracle
-spring.datasource.url: 'jdbc:tc:oracle:slim-faststart:///XEPDB1'
+spring.datasource:
+  url: 'jdbc:tc:oracle:slim-faststart:///XEPDB1'
+  jpa.hibernate.ddl-auto: create
```
pmcphee77, line 79: Do we need the `ddl-auto: create` ?

pmcphee77 (changes requested):

projects/court-case-and-delius/src/main/resources/application.yml
marcus-bcl: Good spot, no idea where that came from - removed!

pmcphee77 (approved):

PR #3893, opened 2024-06-12
PI-2272 throw ignorable message when requirement manager not found.  …

achimber-moj:
…Due to record being soft deleted in delius.

### DIFF
--- a/projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AllocateRequirementService.kt
+++ b/projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AllocateRequirementService.kt
@@ -59,7 +59,7 @@ class AllocateRequirementService(
                 allocationDetail.requirementId,
                 allocationDetail.createdDate
             ) ?: throw IgnorableMessageException(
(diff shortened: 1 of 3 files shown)


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AllocateRequirementService.kt
```diff
@@ -37,7 +37,10 @@ class AllocateRequirementService(
     fun createRequirementAllocation(crn: String, allocationDetail: RequirementAllocation) =
         audit(BusinessInteractionCode.CREATE_COMPONENT_TRANSFER) { audit ->
             val requirement = requirementRepository.findByIdOrNull(allocationDetail.requirementId)
-                ?: throw NotFoundException("Requirement", "id", allocationDetail.requirementId)
+                ?: throw IgnorableMessageException(
+                    "Requirement not found, or soft deleted in delius",
```
marcus-bcl, line 41:
```suggestion
                    "Requirement no longer exists",
```

marcus-bcl (dismissed):

marcus-bcl (approved):

marcus-bcl (approved):

PR #3894, opened 2024-06-12
PI-2253: Added alert contact and description

pmcphee77:


projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseNoteService.kt
```diff
@@ -17,11 +18,14 @@ class CaseNoteService(
...
     private val assignmentService: AssignmentService,
+    private val alertRepository: AlertRepository,
+    private val personManagerRepository: PersonManagerRepository
 ) : AuditableService(auditedInteractionService) {
-
+    @Transactional
     fun createContact(crn: String, createContact: CreateContact) = audit(BusinessInteractionCode.ADD_CONTACT) {
         val person = personRepository.getByCrn(crn)
         val type = caseNoteTypeRepository.getCode(createContact.type.code)
+        val cm = lazy { personManagerRepository.getByCrn(crn) }
```
marcus-bcl, line 28: Why does this need to be lazy?  We're always going to get the value anyway right?

marcus-bcl (approved):

projects/resettlement-passport-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseNoteService.kt
pmcphee77: fair do's

pmcphee77: removed

marcus-bcl (approved):

PR #3905, opened 2024-06-14
PI-2235: Process SENTENCE_CHANGED events with crn

pmcphee77:


projects/custody-key-dates-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTest.kt
```diff
@@ -83,6 +83,44 @@ internal class IntegrationTest {
...
 
+    @Test
+    fun `Custody Key Dates updated from SENTENCE_CHANGED event`() {
+        val notification = Notification(
+            message = MessageGenerator.SENTENCE_CHANGED,
+            attributes = MessageAttributes(eventType = "SENTENCE_CHANGED")
+        )
+
+        val first = CompletableFuture.runAsync {
+            channelManager.getChannel(queueName).publishAndWait(notification, Duration.ofMinutes(3))
```
marcus-bcl, line 94: Do we still need the 3 minute wait here?

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/CustodyDateUpdateService.kt
```diff
@@ -33,6 +34,10 @@ class CustodyDateUpdateService(
         }
     }
 
+    fun findNomsByCrn(crn: String): String {
+        return personRepository.getPersonWithNomsId(crn).nomsId!!
```
marcus-bcl, line 38: This will fail if a CRN doesn't have a corresponding NOMS ID - which will happen a lot.  In that case we should ignore the message, as there'll be no key dates to update.

Also might be worth making this a JPA query repository method (e.g. `"select p.nomsId from Person p where crn = :crn"`), so you're not pulling back the whole person entity.

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -24,11 +24,18 @@ class Handler(
...
+                -> cduService.updateCustodyKeyDates(cduService.findNomsByCrn(message.crn))
+
+                else -> throw IllegalArgumentException("Unexpected offender event type: ${notification.eventType}")
+            }
         }
     }
 }
 
 data class CustodyDateChanged(val bookingId: Long)
+data class ProbationOffenderEvent(val crn: String)
```
marcus-bcl, line 38: Could we annotate these with the `@Message` class.  And also add the new event type to the `@Publish` annotation on the handler method - so it updates the AsyncAPI doc.

You might need to merge from main to get the AsyncAPI stuff.

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -24,11 +24,18 @@ class Handler(
                 ?.let { cduService.updateCustodyKeyDates(it) }
 
             is CustodyDateChanged -> cduService.updateCustodyKeyDates(message.bookingId)
+            is ProbationOffenderEvent -> when (notification.eventType) {
+                "SENTENCE_CHANGED",
+                -> cduService.updateCustodyKeyDates(cduService.findNomsByCrn(message.crn))
```
marcus-bcl, line 29: So if you did turn the NOMS number lookup into a JPA query with a nullable response, you could change this to something like:
```suggestion
                -> cduService.findNomsByCrn(message.crn)?.let { cduService.updateCustodyKeyDates(it) }
```

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/CustodyDateUpdateService.kt
pmcphee77: Im throwing a not found exception in the getPersonWithNomsId when a record isnt found. I thought that this would be better than handling in the service

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
pmcphee77: ok ill rebase

projects/custody-key-dates-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTest.kt
pmcphee77: removed

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/CustodyDateUpdateService.kt
pmcphee77: changed

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
pmcphee77: done

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
pmcphee77: done

marcus-bcl (approved):

PR #3906, opened 2024-06-14
PI-2257 make recall decisions and delius not handling merged crn cases

achimber-moj:


projects/make-recall-decisions-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/recommendation/person/PersonRepository.kt
```diff
@@ -7,7 +7,22 @@ import uk.gov.justice.digital.hmpps.integrations.delius.recommendation.person.en
...
+        WHERE p.crn = :crn AND p.softDeleted = false 
+        AND p.manager.active = true 
+        UNION 
+        SELECT p FROM Person p 
+        LEFT JOIN p.manager
+        JOIN p.additionalIdentifier ai
+        WHERE ai.crn = :crn 
+        AND p.softDeleted = false 
+        AND p.manager.active = true 
+        AND ai.mergeDetail.code IN ('DOFF', 'MFCRN', 'MTCRN', 'PCRN')
```
marcus-bcl, line 23: We only need to check for `'MFCRN'` here.

It might also be a good idea to ensure we don't return multiple records.  For example - if one Person has the crn, and another has the MFCRN additional_identifier, then we should only pull back the first one.

Maybe we could separate them into two repository methods?  Something like:
```kotlin
	@Query("select p from Person p where p.crn = :crn ...")
	fun findByCrn(crn: String): Person?
	
	@Query("select p from Person p left join p.additionalIdentifier ai where ai.type.code = :crn ...")
	fun findByMergedFromCrn(crn: String): Person?
}

fun PersonRepository.getByCrn(crn: String) = findByCrn(crn) ?: findByMergedFromCrn(crn) ?: throw NotFound...
```

marcus-bcl (commented):
Looks good - just a minor comment

projects/make-recall-decisions-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/recommendation/person/PersonRepository.kt
achimber-moj: I will remove the other status values, but the query will only return one value.  I did consider separating into two queries, but one looked like it did the job.  Happy to separate if this is your preference.

marcus-bcl (approved):

PR #3908, opened 2024-06-17
PI-2248: Add sentence changed events to manage pom cases

pmcphee77:


projects/manage-pom-cases-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/PomCaseMessageHandler.kt
```diff
@@ -15,23 +22,76 @@ import uk.gov.justice.digital.hmpps.telemetry.notificationReceived
...
     private val pomAllocated: PomAllocated,
-    private val telemetryService: TelemetryService
-) : NotificationHandler<HmppsDomainEvent> {
+    private val telemetryService: TelemetryService,
+    private val personRepository: PersonRepository
+) : NotificationHandler<Any> {
     @Publish(
         messages = [
             Message(name = "offender-management/handover-changed"),
             Message(name = "offender-management/pom-allocated")
```
marcus-bcl, line 35: Could we add the AsyncAPI doc for the new message type?
```suggestion
            Message(name = "offender-management/pom-allocated"),
            Message(messageId = "SENTENCE_CHANGED", payload = Schema(ProbationOffenderEvent::class)),
```

marcus-bcl (approved):

projects/manage-pom-cases-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/PomCaseMessageHandler.kt
pmcphee77: ahh sorry about that!

marcus-bcl (approved):

PR #3940, opened 2024-06-24
PI-2095: Added all offender managers endpoint

pmcphee77:


projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/Person.kt
```diff
@@ -176,7 +172,15 @@ class Person(
     val restrictionMessage: String? = null,
 
     @Column(columnDefinition = "number")
-    val softDeleted: Boolean = false
+    val softDeleted: Boolean = false,
+
+    @OneToMany(mappedBy = "person")
+    @SQLRestriction("active_flag = 1 and soft_deleted != 1")
```
marcus-bcl, line 178: Should this use `<>`?
```suggestion
    @SQLRestriction("active_flag = 1 and soft_deleted <> 1")
```

Or `= 0` may be simpler?
```suggestion
    @SQLRestriction("active_flag = 1 and soft_deleted = 0")
```

projects/court-case-and-delius/build.gradle.kts
```diff
@@ -13,13 +13,15 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-data-jpa")
     implementation("org.springframework.boot:spring-boot-starter-security")
     implementation("org.springframework.boot:spring-boot-starter-validation")
+    implementation("org.springframework.boot:spring-boot-starter-data-ldap")
```
marcus-bcl, line 16: You'll need to add references to the LDAP creds in the Helm values.yaml file for this too.

pmcphee77: done

marcus-bcl (approved):

PR #3953, opened 2024-06-26
PI-2266: Extra staff details for ap

pmcphee77:


achimber-moj (approved):

projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseDetail.kt
```diff
@@ -38,7 +38,14 @@ data class Profile(
 
 data class Manager(val team: Team)
 
-data class Team(val code: String, val name: String, val ldu: Ldu)
+data class Team(
+    val code: String, val name: String, val ldu: Ldu,
+    val borough: Borough?,
```
marcus-bcl, line 43: Doesn't matter too much, but I wonder if we should avoid the legacy terminology such as probationArea / borough / ldu in the response field names.  These are now called provider / probation delivery unit / local admin unit in Delius - https://ministryofjustice.github.io/hmpps-probation-integration-services/tech-docs/concepts.html#provider

marcus-bcl (approved):
Minor comment - otherwise looks good.

PR #3955, opened 2024-06-26
PI-2105 - add missing field

achimber-moj:


projects/court-case-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/generator/SentenceGenerator.kt
```diff
@@ -19,7 +19,7 @@ import java.time.ZonedDateTime
 
 object SentenceGenerator {
 
-    private val TIME_ZONE = ZoneId.of("Europe/London")
+    val TIME_ZONE = ZoneId.of("Europe/London")
```
marcus-bcl, line 22: You should be able to use `EuropeLondon` from the commons lib for this - see https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/22542d02b4b99ac93af464def13769480ea8e0ef/libs/commons/src/main/kotlin/uk/gov/justice/digital/hmpps/datetime/DeliusDateTimeFormatter.kt#L9

marcus-bcl (approved):

projects/court-case-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/generator/SentenceGenerator.kt
achimber-moj: I will give this a go

achimber-moj: done

marcus-bcl (approved):

PR #3964, opened 2024-07-02
PI-2236: Added bulk update run for custody key dates

pmcphee77:


projects/custody-key-dates-and-delius/src/main/resources/application.yml
```diff
@@ -57,7 +57,9 @@ seed.database: true
 wiremock.enabled: true
 context.initializer.classes: uk.gov.justice.digital.hmpps.wiremock.WireMockInitialiser
 
-messaging.consumer.queue: message-queue
+messaging:
+  consumer.queue: message-queue
+  producer.queue: message-queue
```
marcus-bcl, line 62: Do we need to set MESSAGING_PRODUCER_QUEUE in the values files as well?

marcus-bcl (dismissed):
Couple of minor things - looks good though

marcus-bcl (approved):

PR #3972, opened 2024-07-02
Pi 2321 ignore missing events

pmcphee77:


projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -21,14 +23,27 @@ class Handler(
...
-                ?.let { assessmentSubmitted.assessmentSubmitted(it.crn, it.assessments.first()) }
+        try {
+            if (notification.message.eventType == AssessmentSummaryProduced && featureFlags.enabled("assessment-summary-produced")) {
+                notification.message.detailUrl
+                    ?.let { ordsClient.getAssessmentSummary(URI.create(it)) }
+                    ?.let { assessmentSubmitted.assessmentSubmitted(it.crn, it.assessments.first()) }
+            }
+        } catch (ime: IgnorableMessageException) {
+            telemetryService.trackEvent(
+                "AssessmentSummaryFailure",
```
marcus-bcl, line 39: Could we change this to something like ignored, rather than failed?

```suggestion
                "AssessmentSummaryIgnored",
```

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #3982, opened 2024-07-03
PI-2314: Only process cases with single active custodial events

pmcphee77:


projects/manage-pom-cases-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/Person.kt
```diff
@@ -92,8 +92,21 @@ interface PersonRepository : JpaRepository<Person, Long> {
...
-    @Query("select p.nomsId from Person p where p.crn = :crn and p.softDeleted = false")
-    fun findNomsIdByCrn(crn: String): String?
+    @Query(
+        """
+            select p.noms_number 
+            from offender p
+            join event e on e.offender_id = p.offender_id and e.active_flag = 1 and e.soft_deleted = 0
+            join disposal d on d.event_id = e.event_id and d.active_flag = 1 and d.soft_deleted = 0
+            join custody c on c.disposal_id = d.disposal_id and c.prisoner_number is not null and c.soft_deleted = 0
+            join r_standard_reference_list cs on cs.standard_reference_list_id = c.custodial_status_id and cs.code_value <> 'P'
```
marcus-bcl, line 102: This service doesn't use the custody.prisoner_number or the custodial status - we just need to make sure there's a singe custody record regardless of status.

pmcphee77: Changed

marcus-bcl (approved):

PR #3993, opened 2024-07-05
PI-2306 populate licence conditions for cases with multiple custodial events

achimber-moj:


projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/LicenceConditionApplier.kt
```diff
@@ -65,13 +65,42 @@ class LicenceConditionApplier(
...
+                ),
+                activatedLicence,
+                occurredAt
             )
         }
+
+        //calculated end date should always be populated
+        sentences
+            .maxBy { it.disposal.endDate }.let {
+                return applyLicenceConditions(
```
marcus-bcl, line 98: Is this potentially adding the licence conditions to two of the events?  i.e. the one with the latest `enteredSenetenceEndDate`, and then the one with the latest `endDate`?

projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/LicenceConditionApplier.kt
```diff
@@ -65,13 +65,42 @@ class LicenceConditionApplier(
...
+    }
+
+    fun associateLicenceToLongestSentence(
+        sentences: List<Custody>,
+        com: PersonManager,
+        activatedLicence: ActivatedLicence,
+        occurredAt: ZonedDateTime,
+        properties: Map<String, String>
+    ): List<ActionResult> {
+        val longestSentenceByEnteredDate = sentences
```
marcus-bcl, line 79: Does longest mean the latest by end date here?  Or are we assuming the start date is the same on all matching sentences?

achimber-moj: i mean latest by end date.  i will rename the method to make it clearer.

projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/LicenceConditionApplier.kt
achimber-moj: it should not to, as line 84 should return if the condition is met.

marcus-bcl (approved):

PR #4007, opened 2024-07-09
Feature/pi 2249 bulk update handover dates

achimber-moj:


projects/manage-pom-cases-and-delius/src/main/resources/application.yml
```diff
@@ -61,7 +61,9 @@ seed.database: true
 wiremock.enabled: true
 context.initializer.classes: uk.gov.justice.digital.hmpps.wiremock.WireMockInitialiser
 
-messaging.consumer.queue: message-queue
+messaging:
+  consumer.queue: message-queue
+  producer.queue: message-queue
```
pmcphee77, line 66: I think      ` MESSAGING_PRODUCER_QUEUE: QUEUE_NAME` will need to be put in values.yaml

pmcphee77 (changes requested):

projects/manage-pom-cases-and-delius/src/main/resources/application.yml
achimber-moj: done

pmcphee77 (approved):

PR #4021, opened 2024-07-11
PI-2333: Handle not found exceptions and add to AssessmentSummaryFail…

pmcphee77:
…ureReport

### DIFF
--- a/projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AssessmentService.kt
+++ b/projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AssessmentService.kt
@@ -27,7 +27,7 @@ class AssessmentService(
     private val eventRepository: EventRepository,
     private val contactService: ContactService
 ) {
-    fun recordAssessment(person: Person, summary: AssessmentSummary, telemetryParams: Map<String, String>) {
+    fun recordAssessment(person: Person, summary: AssessmentSummary) {
         val previousAssessment = oasysAssessmentRepository.findByOasysId(summary.assessmentPk.toString())
 
         val eventNumber = summary.furtherInformation.cmsEventNumber?.toString()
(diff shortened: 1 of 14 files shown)


projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AssessmentSubmitted.kt
```diff
@@ -49,11 +51,16 @@ class AssessmentSubmitted(
             domainEventService.publishEvents(regEvents)
             telemetryService.trackEvent("AssessmentSummarySuccess", telemetryParams)
         } catch (exception: Exception) {
-            telemetryService.trackEvent(
-                "AssessmentSummaryFailure",
-                telemetryParams + ("exception" to exception.message!!)
-            )
-            throw exception
+            when (exception) {
+                is NotFoundException -> throw IgnorableMessageException(exception.message!!, telemetryParams)
```
marcus-bcl, line 55: Would it be better to be more explicit about which NotFoundExceptions we're ignoring?  i.e. update the Event and Person repositories to throw an IgnorableMessageException instead.

Also, maybe we should remove this try-catch altogether from here - all it's doing is adding a custom event for every exception, but we can see the exceptions in App Insights anyway so I don't think we need the extra event.

projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AssessmentService.kt
```diff
@@ -27,7 +27,7 @@ class AssessmentService(
     private val eventRepository: EventRepository,
     private val contactService: ContactService
 ) {
-    fun recordAssessment(person: Person, summary: AssessmentSummary) {
+    fun recordAssessment(person: Person, summary: AssessmentSummary, telemetryParams: Map<String, String>) {
```
marcus-bcl, line 30: Is `telemetryParams` being used anywhere?

pmcphee77: was going to but decided not to. Removed

marcus-bcl (approved):

PR #4038, opened 2024-07-16
PI-2358 custody key dates update do not ignore soft deleted records

achimber-moj:


projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/KeyDate.kt
```diff
@@ -48,4 +42,6 @@ class KeyDate(
             it.version = version
```
marcus-bcl, line 42: Don't we need to un-delete the record as well?
```suggestion
            it.version = version
            it.softDeleted = false
```

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/KeyDate.kt
```diff
@@ -48,4 +42,6 @@ class KeyDate(
...
         }
     }
+
+    fun getSoftDeleteFlag() = softDeleted
```
marcus-bcl, line 46: Do you need this getter?

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/KeyDate.kt
achimber-moj: the entity set the soft delete flag to false by default.  however, i will make the change to make this clearer

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/KeyDate.kt
achimber-moj: i was using it in my test, but will delete it, i am can use the property directly

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/KeyDate.kt
achimber-moj: done

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/KeyDate.kt
achimber-moj: done

marcus-bcl (approved):

PR #4055, opened 2024-07-19
Pi 2365 add compare

pmcphee77:
Examples below
image
image


projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -57,7 +59,68 @@ class CommunityApiController(
...
                 .headers(ex.responseHeaders)
-                .body(ex.responseBodyAsString);
+                .body(ex.responseBodyAsString)
+        }
+    }
+
+    @PostMapping("/compare")
+    fun compare(@RequestBody compare: Compare, request: HttpServletRequest): CompareReport {
+        val headers = request.headerNames.asSequence().associateWith { request.getHeader(it) }
+            .filter { it.key.lowercase() == "Authorization".lowercase() }.toMutableMap()
```
marcus-bcl, line 69: Is this just getting the auth header?  Could it be simplified to something like:

```suggestion
        val headers = mapOf("Authorization" to request.getHeader("Authorization"))
```

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -57,7 +59,68 @@ class CommunityApiController(
...
+                "Values differ at ${it.asJsonObject()["path"].toString()} \n" +
+                    "Found $ccdApiValue in new API, but is ${it.asJsonObject()["value"].toString()} in Community API"
+            } else if (op.contains("remove")) {
+                "Additional element at $path exists in new API but is not present in Community API"
+            } else if (op.contains("add")) {
+                "Element at $path not found in new API, but is present with value of ${it.asJsonObject()["value"].toString()} in Community API "
+            } else {
+                "Unhandled operation $op"
+            }
+        }.map { x -> x.value }.toList()
```
marcus-bcl, line 108: I don't think you need the associateWith here, as you're mapping it back to the value?  Could be a simple map:
```suggestion
        val results = diff.toJsonArray().map {
            val op = it.asJsonObject()["op"].toString()
            val path = it.asJsonObject()["path"].toString()
            if (op.contains("replace")) {
                val ccdApiValue = getValueFromJsonPath(ccdJson, path)
                "Values differ at $path \n" +
                    "Found $ccdApiValue in new API, but is ${it.asJsonObject()["value"].toString()} in Community API"
            } else if (op.contains("remove")) {
                "Additional element at $path exists in new API but is not present in Community API"
            } else if (op.contains("add")) {
                "Element at $path not found in new API, but is present with value of ${it.asJsonObject()["value"].toString()} in Community API "
            } else {
                "Unhandled operation $op"
            }
        }
```

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -57,7 +59,68 @@ class CommunityApiController(
...
+    fun compare(@RequestBody compare: Compare, request: HttpServletRequest): CompareReport {
+        val headers = request.headerNames.asSequence().associateWith { request.getHeader(it) }
+            .filter { it.key.lowercase() == "Authorization".lowercase() }.toMutableMap()
+
+        headers.forEach { log.info("Header name: ${it.key}") }
+
+        val uri = try {
+            Uri.valueOf(compare.uri)
+        } catch (ex: Exception) {
+            return compare.toReport("This endpoint is not configured")
```
marcus-bcl, line 76: Is it worth just throwing on these exceptions instead of catching and returning a 200?  Would simplify the code, and means we can see them in Sentry etc.

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -57,7 +59,68 @@ class CommunityApiController(
...
+        val comApiUri = uri.comApiUrl.replace("{crn}", compare.crn)
+        val comApiJsonString = try {
+            communityApiClient.proxy(URI.create(communityApiUrl + comApiUri), headers).body!!
+        } catch (ex: HttpStatusCodeException) {
+            log.error("Exception thrown when calling ${communityApiUrl + comApiUri}. community-api returned ${ex.message}")
+            return compare.toReport(ex.message ?: "No message")
         }
+        val ccdJson = Json.createReader(StringReader(ccdJsonString)).readValue().asJsonObject()
+        val comApiJson = Json.createReader(StringReader(comApiJsonString)).readValue().asJsonObject()
+        val diff: JsonPatch = Json.createDiff(ccdJson, comApiJson)
```
marcus-bcl, line 92: Did you consider using [JSONAssert](https://github.com/skyscreamer/JSONassert#error-messages), or any other JSON comparison library for this?  Happy to continue with the javax dependencies and jsonpatch - just wondering.

settings.gradle.kts
```diff
@@ -91,6 +91,8 @@ dependencyResolutionManagement {
             library("asyncapi", "org.openfolder:kotlin-asyncapi-spring-web:3.0.3")
             library("wiremock", "org.wiremock:wiremock-standalone:3.8.0")
             library("flipt", "io.flipt:flipt-java:1.1.1")
+            library("javax-json", "javax.json:javax.json-api:1.1.2")
+            library("glassfish-json", "org.glassfish:javax.json:1.1.2")
```
marcus-bcl, line 95: Should we be using the jakarta versions these days over javax?

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
pmcphee77: Removed exception handling

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
pmcphee77: fixed during refactor

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
pmcphee77: Fixed

settings.gradle.kts
pmcphee77: Fixed

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
pmcphee77: I found that JsonPatch provided a flexible way of looking at the differences. So just went with that

marcus-bcl (approved):

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
pmcphee77: fixed

PR #4077, opened 2024-07-29
PI-2395 split name into separate fields

achimber-moj:


projects/subject-access-requests-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SubjectAccessRequestsService.kt
```diff
@@ -14,8 +15,13 @@ class SubjectAccessRequestsService(private val personRepository: PersonRepositor
     }
 }
 
-fun PersonEntity.toPerson(): Person = Person(getName())
+fun PersonEntity.toPerson(): Person {
```
pmcphee77, line 18: i dont think you need the `let` here. Just use `listOfNotNull(secondName, thirdName).joinToString(" ")`

achimber-moj: this does not result in a null response, but a "" for the middle name.

pmcphee77 (approved):

PR #4085, opened 2024-07-31
PI-2391: removed soft deleted check from order manages to fix orderin…

pmcphee77:
…g issues


projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -202,18 +202,27 @@ class CommunityApiController(
                 .map(CompletableFuture<CompareReport>::join)
                 .collect(Collectors.toList())
         }
+
+        val executed = reports.filter { it.testExecuted == true }
+        val successful = reports.filter { it.success && it.testExecuted == true }
         val unsuccessful = reports.filter { !it.success && it.testExecuted == true }
```
marcus-bcl, line 208: Minor - but there's another handy function for that in Kotlin 😄 
```suggestion
        val (successful, unsuccessful) = executed.partition { it.success }
```

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -202,18 +202,27 @@ class CommunityApiController(
...
-        val executionFailures = reports.filter { it.testExecuted == false }.size
+        val executionFailures = reports.filter { it.testExecuted == false }
+        val distinctEndpointsExecuted = executed.groupBy(CompareReport::endPointName)
+        val covered = distinctEndpointsExecuted.map { it.key + " covered " + it.value.size + " times" }
+        val requestedButNotCovered = compare.uriConfig.entries.map { it.key }.toSet()
+            .filter { e -> !distinctEndpointsExecuted.map { it.key }.toSet().contains(e) }
```
marcus-bcl, line 213:
```suggestion
        val requestedButNotCovered = compare.uriConfig.keys
            .filter { !distinctEndpointsExecuted.keys.contains(it) }
```

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/service/OffenderService.kt
```diff
@@ -142,7 +142,7 @@ fun Person.toOffenderManagers() = offenderManagers.sortedByDescending { it.date
                 code = team.code,
                 description = team.description,
                 telephone = team.telephone,
-                emailAddress = team.emailAddress,
+                //emailAddress = team.emailAddress,
```
marcus-bcl, line 145: Is this meant to be left commented out?

marcus-bcl (approved):

PR #4115, opened 2024-08-05
PI-2103: Added get by nsiId

pmcphee77:


projects/court-case-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/ProxyIntegrationTest.kt
```diff
@@ -29,6 +30,7 @@ import uk.gov.justice.digital.hmpps.test.MockMvcExtensions.contentAsJson
 import uk.gov.justice.digital.hmpps.test.MockMvcExtensions.withToken
 
 @AutoConfigureMockMvc
+@TestPropertySource(properties = ["lao-access.ignore-exclusions = false", "lao-access.ignore-restrictions = true"])
```
marcus-bcl, line 33: Are we still using these fields, or are we planning to remove LAO in another PR?

projects/court-case-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/ProxyIntegrationTest.kt
```diff
@@ -313,5 +319,3 @@ internal class ProxyIntegrationTest {
         assertThat(res.totalNumberOfRequests, equalTo(1))
     }
 }
-
-//{"status":403,"developerMessage":"This is a restricted record. Please contact a system administrator"}
```
marcus-bcl, line 317: Can this be removed?

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -198,7 +220,8 @@ class CommunityApiController(
         val reports = personList.content.flatMap { person ->
             val convictionId = person.events.filter { it.disposal != null }.maxOfOrNull { it.id }
             val nsiCodes = person.nsis.filter { it.eventId == convictionId }.map { it.type.code.trim() }
-            runAll(person.crn, convictionId, nsiCodes, compare, headers).stream()
+            val nsiId = person.nsis.firstOrNull()?.id
+            runAll(person.crn, convictionId, nsiCodes, nsiId, compare, headers).stream()
                 .map(CompletableFuture<CompareReport>::join)
                 .collect(Collectors.toList())
```
marcus-bcl, line 226: Not really a change in this PR, but just spotted you don't need to convert to a stream here.  This would also work:
```suggestion
            runAll(person.crn, convictionId, nsiCodes, nsiId, compare, headers).map { it.join() }
```

Also I think this is running in sequence and not in parallel?  Think you need to use `CompletableFuture.allOf` to get them to run in parallel:
```suggestion
            CompletableFuture.allOf(person.crn, convictionId, nsiCodes, nsiId, compare, headers).join()
            return listOf(person.crn, convictionId, nsiCodes, nsiId, compare, headers).map { it.get() }
```

projects/court-case-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/ProxyIntegrationTest.kt
pmcphee77: Planning on removing in another PR

projects/court-case-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/ProxyIntegrationTest.kt
pmcphee77: It has been removed - its commented that i have removed

marcus-bcl: Doh sorry I always make that mistake

projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
pmcphee77: Changed

marcus-bcl (approved):

PR #4144, opened 2024-08-07
PI-2101: Added licence conditions endpoint

pmcphee77:


projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -49,7 +49,7 @@ class Handler(
             is CustodyDateChanged -> cduService.updateCustodyKeyDates(message.bookingId)
             is ProbationOffenderEvent -> when (notification.eventType) {
                 "SENTENCE_CHANGED",
-                -> personRepository.findNomsIdByCrn(message.crn)?.let { cduService.updateCustodyKeyDates(it) }
+                    -> personRepository.findNomsIdByCrn(message.crn)?.let { cduService.updateCustodyKeyDates(it) }
```
marcus-bcl, line 52: Maybe IntelliJ updated the formatting rules for this in the latest version? Should be fine to merge with your PR

marcus-bcl (approved):

PR #4167, opened 2024-08-12
PI-2907: Fixed issue with document download endpoint

pmcphee77:


projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/proxy/CommunityApiController.kt
```diff
@@ -343,7 +352,7 @@ class CommunityApiController(
         return proxy(request)
     }
 
-    fun proxy(request: HttpServletRequest): Any {
+    fun proxy(request: HttpServletRequest): ResponseEntity<ByteArray> {
```
marcus-bcl, line 355: Will this load the whole doc into memory as a byte array?  Could cause crashes when downloading large files?

pmcphee77: hmm you are right it will. Ill change it so that the call to community api uses a Stream

marcus-bcl (approved):

PR #4172, opened 2024-08-13
PI-2380: Deploy mms to preprod

pmcphee77:


projects/manage-supervision-and-delius/deploy/values-preprod.yml
```diff
@@ -1,4 +1,4 @@
-enabled: false # TODO set this to true when you're ready to deploy your service
+enabled: true
```
achimber-moj, line 1: should be able to remove this line.  the default value is set to true.

achimber-moj (changes requested):

projects/manage-supervision-and-delius/deploy/values-preprod.yml
pmcphee77: Yep! ive changed both the dev and preprod one

achimber-moj (approved):

PR #4174, opened 2024-08-14
PI-2426 - use new fields to populate a contact

achimber-moj:


projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ReferralService.kt
```diff
@@ -142,7 +142,8 @@ class ReferralService(
         }
         contactService.createContact(
             ContactDetails(
-                date = details.cancelledAt,
+                date = details.cancellationRecordedAt,
+                cancellationRecordedAt = details.cancelledAtDate,
```
pmcphee77, line 146: Should these be the other way round?

pmcphee77 (changes requested):

projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ReferralService.kt
achimber-moj: cancellationRecordedAt is ZonedDateTime ContactDetails.date is also ZonedDateTime.  cancelledAtDate is of type LocalDate.  When create a contact field contact_date is LocalDate and will be populated by cancelledAtDate, contact field contact_start_time will be populated by cancellationRecordedAt.  

see payload extract when calling approved premises 

    "cancelledAtDate": "2024-08-14",
    "cancellationRecordedAt": "2024-08-14T08:47:12.290096228Z"

pmcphee77 (approved):

PR #4199, opened 2024-08-27
PI-2449: Updated pathfinder to be the same as soc

pmcphee77:


projects/pathfinder-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/UserController.kt
marcus-bcl, line 1: We still need this endpoint for pathfinder - they're currently using it in live.  The changes looks good otherwise.

pmcphee77: reinstated this endpoint

marcus-bcl (approved):

projects/pathfinder-and-delius/build.gradle.kts
```diff
@@ -35,4 +35,8 @@ configure<ClassPathExtension> {
         "**/entity/**",
         "**/AppKt.class"
     )
+
+    sonarExclusions = listOf(
+        "**/hmpps/**"
+    )
```
marcus-bcl, line 41: Not sure if we need this to be honest - we can still approve/merge the PR without it.  This might bite us as the Pathfinder/SOC code diverges.

marcus-bcl (approved):

marcus-bcl (approved):

PR #4202, opened 2024-08-28
PI-2457: Added mappa endpoint to MPC

pmcphee77:


projects/manage-pom-cases-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/registration/entity/Registration.kt
```diff
@@ -88,6 +109,16 @@ interface RegistrationRepository : JpaRepository<Registration, Long> {
...
     )
     fun hasVloAssigned(personId: Long): Boolean
+
+    @Query(
+        "select registration from Registration registration " +
+            "where registration.type.code = 'MAPP' " +
+            "and registration.person.id = :offenderId " +
+            "and registration.softDeleted = false " +
+            "and registration.deRegistered = false " +
+            "order by registration.createdDatetime desc"
```
marcus-bcl, line 119: Could we use a multi-line string here?

projects/manage-pom-cases-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/MappaDetail.kt
```diff
@@ -0,0 +1,60 @@
...
+    val team: KeyValue,
+
+    @Schema(description = "Officer")
+    val officer: StaffHuman,
+
+    @Schema(description = "Probation area")
+    val probationArea: KeyValue,
+
+    @Schema(description = "Notes")
+    private val notes: String? = null
```
marcus-bcl, line 43: We can drop most of these fields from the Community API version - MPC aren't using them: https://github.com/ministryofjustice/offender-management-allocation-manager/blob/25bb19b0ad89974898d7fd301ad65a3c4d3b8f6a/app/services/offender_service.rb#L105-L113

marcus-bcl (approved):

PR #4223, opened 2024-09-10
PI-2515: CVL Text has a 4000 byte limit - truncate bytes

pmcphee77:


projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/LicenceConditionService.kt
```diff
@@ -32,7 +32,7 @@ class LicenceConditionService(
                 category,
                 subCategory,
                 notes,
-                cvlText.let { it?.take(4000) }
+                cvlText?.let { String(it.toByteArray().take(4000).toByteArray()) }
```
achimber-moj, line 35: do we need to do a trim as well if there are any leading spaces

achimber-moj (approved):

PR #4251, opened 2024-09-20
PI-2541 multiple pom records are custody status to delius message

achimber-moj:


projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/Person.kt
```diff
@@ -19,10 +20,54 @@ class Person(
...
 
 interface PersonRepository : JpaRepository<Person, Long> {
     fun findByNomsNumberAndSoftDeletedIsFalse(nomsNumber: String): List<Person>
+
+    @Query(
+        """
+        SELECT p FROM Person p 
+        LEFT JOIN p.manager
+        JOIN p.additionalIdentifier ai
+        WHERE ai.id = :id
```
marcus-bcl, line 41: Should this be checking the person's id?
```suggestion
        WHERE ai.person.id = :id 
```

Also may be worth returning whether the count is greater than 0, rather than returning the whole Person graph?

And do you need the active manager check?

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/manager/prison/PrisonManagerService.kt
```diff
@@ -54,8 +58,20 @@ class PrisonManagerService(
...
+        try {
+            activePrisonManager = prisonManagerRepository.findActiveManagerAtDate(person.id, allocationDate)
+        } catch (e: IncorrectResultSizeDataAccessException) {
+            if (personRepository.findByMergedFromCrn(person.id) == null) {
+                throw e
+            }
+            //where crn has a merged from record
+            //we can ignore the IncorrectResultSizeDataAccessException and
+            //process the record no further
+            throw IgnorableMessageException("Person has merged from record", mapOf("person id" to person.id.toString()))
```
marcus-bcl: I think the NOMS number is automatically added to these - so no need to add the person id manually.  Also, the other IgnorableMessageException instances in this service seem to have a slightly different message format - probably worth making this one consistent e.g. something like
```suggestion
            throw IgnorableMessageException("MergedPrisonManagerHistory")
```

achimber-moj: already added

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/Person.kt
achimber-moj: took the query from another project where the same thing is being used, happy to use a count.  if we do not need to check for an active manager i can remove

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/manager/prison/PrisonManagerService.kt
marcus-bcl: What I mean is you don't need to add any additional properties here, the NOMS number is already added by default in the Handler: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/559f1b6a261fc25883c0ed70f1e03919f7b22df5/projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt#L110-L115

Also could we shorten the exception message to be consistent with the others in this project?  e.g. 
```
throw IgnorableMessageException("MergedPrisonManagerHistory")
```

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/Person.kt
marcus-bcl: I don't think we need to check for an active manager, we ideally just want a query that returns true/false if the case has been merged.

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/manager/prison/PrisonManagerService.kt
achimber-moj: the action processor allow additional properties.  how else will will tie the message with the crn if we do not add the identifier.  i am happy with the identifier, and would need further convincing to change

marcus-bcl: Ah! Sorry, not trying to be a pain honestly, but I just had a look through the existing code and realised we can't use an IgnorableMessageException here anyway 🤦 as it will interrupt the UpdateLocationAction and means it won't create the location change contact.  So we need to change this to just `return` anyway!

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/entity/Person.kt
marcus-bcl: The updated code is still checking the additional_identifier_id instead of the offender_id column, and still joining on manager.  There are also a few unnecessarily mapped columns in the AdditionalIdentifier entity - you can simplify the whole thing quite a bit... this should work:

```kotlin

    @Query(
        """
        select count(ai) > 0 
        from AdditionalIdentifier ai
        where ai.personId = :personId
        and ai.name.code = 'MFCRN'             
        """
    )
    fun personHasBeenMerged(personId: Long): Boolean
}

@Entity
@Immutable
@SQLRestriction("soft_deleted = 0")
class AdditionalIdentifier(
    @Id
    @Column(name = "additional_identifier_id")
    val id: Long,

    @Column(name = "offender_id")
    val personId: Long,

    @ManyToOne
    @JoinColumn(name = "identifier_name_id")
    val name: ReferenceData,

    @Column(updatable = false, columnDefinition = "number")
    val softDeleted: Boolean = false,
)

```

projects/prison-custody-status-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/person/manager/prison/PrisonManagerService.kt
achimber-moj: would we not already have created a updated contact from the pre merge message.  why do we need to create a contact for the location change?

marcus-bcl (approved):

PR #4257, opened 2024-09-23
MAN-26: Added caseload search filter and sort

pmcphee77:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
```diff
@@ -218,20 +223,194 @@ data class Caseload(
...
+        join fetch c.team t
+        left join fetch c.nextAppointment na
+        left join fetch na.type naType
+        left join fetch c.previousAppointment pa
+        left join fetch pa.type paType
+        left join fetch c.latestSentence ls
+        left join fetch ls.mainOffence mo
+        left join fetch mo.offence moo
+        where c.staff.code = :staffCode
+        and ((:nameOrCrn is null or upper(p.crn) like '%' || upper(:nameOrCrn) || '%' or upper(p.forename || ' ' || p.surname) like '%' || upper(:nameOrCrn) || '%'))
```
marcus-bcl, line 401: Should this include a filter on surname+forename too?  E.g. if someone searched for "Smith, John"?
```suggestion
        and (
          :nameOrCrn is null 
          or upper(p.crn) like '%' || upper(:nameOrCrn) || '%' 
          or upper(p.forename || ' ' || p.surname) like '%' || upper(:nameOrCrn) || '%'
          or upper(p.surname || ' ' || p.forename) like '%' || upper(:nameOrCrn) || '%'
          or upper(p.surname || ', ' || p.forename) like '%' || upper(:nameOrCrn) || '%')
```

Feel free to ignore this if you think it makes the query too verbose though.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/controller/CaseloadController.kt
```diff
@@ -37,4 +51,17 @@ class CaseloadController(private val userService: UserService) {
     @GetMapping("/team/{teamCode}/staff")
     @Operation(summary = "Gets the staff within the team")
     fun getTeamStaff(@PathVariable teamCode: String) = userService.getTeamStaff(teamCode)
+
+    private fun sort(sortString: String): Sort {
+
+        val regex = Regex(pattern = "[A-Za-z0-9]+\\.(ASC|DESC)", options = setOf(RegexOption.IGNORE_CASE))
```
marcus-bcl, line 57: nitpick (feel free to ignore) - do you need A-Za-z when the regex is case-insensitive?
```suggestion
        val regex = Regex(pattern = "[A-Z0-9]+\\.(ASC|DESC)", options = setOf(RegexOption.IGNORE_CASE))
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
```diff
@@ -218,20 +223,194 @@ data class Caseload(
...
+            select e.*,
+            row_number() over (partition by e.offender_id order by cast(e.event_number as number) desc) as row_num 
+            from event e join main_offence mo on mo.event_id = e.event_id
+            where e.soft_deleted = 0 
+            and e.active_flag = 1
+            ) sub
+        where sub.row_num = 1
+"""
+)
+@Synchronize("event")
```
marcus-bcl, line 261: Nice - I've not really used `@Subselect` + `@Synchronize` before. Looks clever 😄

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
```diff
@@ -218,20 +223,194 @@ data class Caseload(
...
+@Subselect(
+    """
+        select * from (
+            select 
+            q.staff_id,
+            q.offender_id,
+            q.appointment_datetime,
+            q.contact_id,
+            q.contact_type_id,
+            row_number() over (partition by staff_id, offender_id order by appointment_datetime asc) as row_num
```
marcus-bcl, line 283: Why do we need to partition by staff id here?  Are we only showing the latest/next appointment for the current user with the person on probation?

marcus-bcl (commented):
Looks good!  Just added a couple of questions

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
pmcphee77: we are actually looking for the latest appointment for the staff record here as we are looking for the staffs appointment with the crn. The appointment for the crn could in theory be with another staff member

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
pmcphee77: agreed! i like it. Thanks. Done

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/controller/CaseloadController.kt
pmcphee77: Yep - dont need that. Changed

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
pmcphee77: Yeah handy for these situations

marcus-bcl (approved):

PR #4319, opened 2024-10-16
PI-2575

joseph-bcl:
- Create common-platform-and-delius service


projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Notifier.kt
```diff
@@ -0,0 +1,63 @@
...
+import uk.gov.justice.digital.hmpps.publisher.NotificationPublisher
+
+@Service
+@Channel("hmpps-domain-events-topic")
+class Notifier(
+    @Qualifier("topicPublisher") private val topicPublisher: NotificationPublisher,
+) {
+    @Subscribe(
+        messages = [
+            Message(title = "probation-case.engagement.created", payload = Schema(HmppsDomainEvent::class))
```
marcus-bcl, line 21: Could you add another line in here for the address domain event?

projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/PersonAddress.kt
```diff
@@ -0,0 +1,112 @@
...
+    val person: Person,
+
+    @Column(name = "notes", columnDefinition = "clob")
+    val notes: String? = null,
+
+    @Column(name = "address_number")
+    val addressNumber: String? = null,
+
+    @Column(name = "street_name")
+    var streetName: String? = null,
```
marcus-bcl, line 61: Any reason not to use `val` here, given we're not updating addresses in place?

projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -1,29 +1,82 @@
...
+        val defendants = notification.message.hearing.prosecutionCases
+            .flatMap { it.defendants }
+            .filter { defendant ->
+                defendant.offences.any { offence ->
+                    offence.judicialResults.any { judicialResult ->
+                        judicialResult.isConvictedResult == true && judicialResult.label == "Remanded in custody"
+                    }
+                }
+            }
+            .ifEmpty { throw IllegalArgumentException("No valid defendants found") }
```
marcus-bcl, line 40: Do we need to throw an exception here?

PR #4321, opened 2024-10-16
Man 28 sentence license condition notes

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceService.kt
```diff
@@ -67,9 +69,57 @@ class SentenceService(
...
+                    note.let { n ->
+                        when {
+                            n.length > noteLength -> true
+                            else -> false
+                        }
+                    }
+                )
+            }
+
+        }
```
marcus-bcl, line 110: Not 100% sure what this is trying to do?  Is it to move the "Comment added by ..." line out?

achimber-moj: yes

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceService.kt
```diff
@@ -91,13 +92,21 @@ class SentenceService(
...
+            return notes.split(splitParam).map { note ->
+                val addedBy = Regex(
+                    "^Comment added by .+? on \\d{2}/\\d{2}/\\d{4} at \\d{2}:\\d{2}"
+                        + System.lineSeparator()
+                ).find(note)?.value
+                val noteText = addedBy?.let { note.removePrefix(addedBy) } ?: note
+
+                val createdBy = addedBy?.removeSuffix(System.lineSeparator())?.reversed()?.substring(23)?.reversed()
+                val dateCreatedBy =
+                    addedBy?.removeSuffix(System.lineSeparator())?.reversed()?.substring(9, 19)?.reversed()
```
marcus-bcl, line 104: Might be easier adding capturing groups to the regex to grab the bits you need?  Something like:
```kotlin
                val matchResult = Regex(
                    "^Comment added by (.+?) on (\\d{2}/\\d{2}/\\d{4}) at \\d{2}:\\d{2}"
                        + System.lineSeparator()
                ).find(note)
                val commentLine = matchResult?.value
                val commentText = commentLine?.let { note.removePrefix(commentLine) } ?: note

                val userCreatedBy = matchResult?.groupValues[1]
                val dateCreatedBy = matchResult?.groupValues[2]?.let { DeliusDateFormatter.parse(it) }
                
                LicenceConditionNote(
                    userCreatedBy,
                    dateCreatedBy,
                    commentText,
                    note.length > noteLength
                )
```

If you wanted to stick with the current approach, you could look at Kotlin's [take](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/take.html) and [takeLast](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/take-last.html), which is useful for grabbing prefixes and suffixes of a fixed length.

marcus-bcl (approved):

PR #4355, opened 2024-10-29
PI-2575 - Change hearing message fields to be nullable

joseph-bcl:
- Added validation to prevent the creation of records if a defendants' firstname, surname or dob is null


projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -44,7 +44,9 @@ class Handler(
         val courtCode = notification.message.hearing.courtCentre.code
 
         defendants.forEach { defendant ->
-            val matchRequest = defendant.toProbationMatchRequest()
+
+            val matchRequest = defendant.toProbationMatchRequest() ?: return
```
marcus-bcl, line 48: This will return out of the handler altogether - would we rather continue looping in case there are other valid defendants?

You can use [return labels](https://kotlinlang.org/docs/returns.html#return-to-labels) in Kotlin to return to the forEach loop e.g.
```suggestion
            val matchRequest = defendant.toProbationMatchRequest() ?: return@forEach
```

Should probably do the same for the one further down too.

marcus-bcl (approved):

PR #4365, opened 2024-10-30
Man 109 backend create appointment in mms

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Contact.kt
```diff
@@ -98,6 +99,11 @@ class Contact(
     @Column(name = "soft_deleted", columnDefinition = "NUMBER", nullable = false)
     val softDeleted: Boolean = false
 ) {
+    var partitionAreaId: Long = 0
+
+    @CreatedBy
```
marcus-bcl, line 104: This annotation won't do anything without the auditing entity listener annotation, probably worth removing it if you only need to read the createdByUserId - also can these two fields be `val`, or do we change them somewhere?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/appointment/CreateAppointment.kt
```diff
@@ -0,0 +1,32 @@
...
+    val type: Type,
+    val start: ZonedDateTime,
+    val end: ZonedDateTime?,
+    val interval: Int,
+    val eventId: Long,
+    val requirementId: Long? = null,
+    val licenceConditionId: Long? = null,
+    val numberOfAppointments: Int? = null,
+    val until: ZonedDateTime? = null,
+    val uuid: UUID = UUID.randomUUID()
```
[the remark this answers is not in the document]
achimber-moj: i have updated the code so that uuid is not preset, as this api will be used for the f/e

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Contact.kt
achimber-moj: updated

marcus-bcl (approved):

PR #4367, opened 2024-10-30
CAS-1178 - setup documents data for cas1 team

gregkhawkins:
**Background**
* We are migrating away from `Community API`'s `GET secure/offenders/crn/{crn}/documents/grouped` endpoint
* Instead we plan to integrate with the `AP and Delius API`'s `GET /documents/{crn}/all` endpoint
* For everything to work locally in our `AP tools` (i.e. local docker infrastructure) we need some extra documents to be setup in the `AP and Delius API`'s local H2 database

**PR includes**
- Data setup for new docs
- New wiremock mapping file to surface docs in the way we require


marcus-bcl (approved):

marcus-bcl:
@gregkhawkins we enforce signed commits in this repo.  Happy to help you set that up - you just need to generate a GPG key, register it with your GitHub account, and use it to sign your commits.  See https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

marcus-bcl (approved):

PR #4381, opened 2024-11-01
Man 109 schedule multiple appointments

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceAppointmentService.kt
```diff
@@ -29,16 +33,43 @@ class AppointmentService(
...
-            val appointment = appointmentRepository.save(createAppointment.withManager(om))
-            val createdAppointment = CreatedAppointment(appointment.id)
-            audit["contactId"] = appointment.id
+            val createAppointments: ArrayList<CreateAppointment> = arrayListOf()
 
-            return@audit createdAppointment
+            createAppointment.let {
+                val numberOfAppointments = createAppointment.until?.let {
+                    Period.between(createAppointment.start.toLocalDate(), it.toLocalDate()).days
+                } ?: createAppointment.numberOfAppointments
```
marcus-bcl, line 46: I'm probably missing something here, how does this account for the interval?

E.g. if you had:
```
start=01/11/2024
until=01/12/2024
interval=7 days
```
this calculation would give you `numberOfAppointments=30`, but shouldn't it be 4?

achimber-moj: updated

marcus-bcl (approved):

PR #4393, opened 2024-11-05
MAN-109 - Use Duration instead of Period when calculating number of a…

achimber-moj:
…ppointments. Amend exception type so that message is returned via controller advice

### DIFF
--- a/projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceAppointmentService.kt
+++ b/projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceAppointmentService.kt
@@ -1,8 +1,6 @@
 package uk.gov.justice.digital.hmpps.service
 
-import org.springframework.http.HttpStatus
 import org.springframework.stereotype.Service
-import org.springframework.web.server.ResponseStatusException
 import uk.gov.justice.digital.hmpps.api.model.appointment.AppointmentDetail
 import uk.gov.justice.digital.hmpps.api.model.appointment.CreateAppointment
 import uk.gov.justice.digital.hmpps.api.model.appointment.CreatedAppointment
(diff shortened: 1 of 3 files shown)


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceAppointmentService.kt
```diff
@@ -45,10 +44,12 @@ class SentenceAppointmentService(
...
-                    Period.between(
-                        createAppointment.start.toLocalDate(),
-                        it.toLocalDate()
-                    ).days.div(createAppointment.interval.value)
+                    val duration = Duration.between(
+                        createAppointment.start.toLocalDateTime(),
+                        it.toLocalDateTime()
+                    ).toDays() + 1
+
+                    ceil(duration.toDouble().div(createAppointment.interval.value)).toInt()
```
marcus-bcl, line 52: Why do you need the ceil / toDouble here, and the plus 1 above? If you have a duration of 30 days and an interval of 7 days, should there be 4 appointments or 5? I may be misunderstanding, but I'd expect 4 - if so, removing the ceil and the plus 1 gives the right answer.  Example here: https://pl.kotl.in/CeB_t339V
```suggestion
                    val duration = Duration.between(
                        createAppointment.start.toLocalDateTime(),
                        it.toLocalDateTime()
                    ).toDays()
                    (duration / createAppointment.interval.value).toInt()
```

achimber-moj: it is matching delius functionality.

marcus-bcl: Ah okay makes sense now - because Delius is creating one at the start time as well as at each interval.  Might still be worth simplifying this then, to just add 1 at the end?
```suggestion
                    val duration = Duration.between(
                        createAppointment.start.toLocalDateTime(),
                        it.toLocalDateTime()
                    ).toDays()

                    (duration / createAppointment.interval.value).toInt() + 1
```

marcus-bcl (approved):

PR #4442, opened 2024-11-19
Man 108 record if complied and update note

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AppointmentOutcomeService.kt
```diff
@@ -21,8 +23,17 @@ class AppointmentOutcomeService(
...
         appointment.apply {
+            val dateTime = LocalDateTime.now()
             attended = outcome.attended
-            notes = outcome.notes
+            complied = if (contactTypeOutcome.outcome.outcomeCompliantAcceptable!!) "Y" else "N"
+            notes = outcome.notes?.let {
+                """
+                Comment added by ${outcome.recordedBy} on ${
+                    dateTime.format(DeliusDateTimeFormatter).substring(0, 10)
+                } at ${dateTime.format(DeliusDateTimeFormatter).substring(11, 16)}
```
marcus-bcl, line 33: Isn't this line added by the Delius DB trigger?  Do you need to add it manually?

achimber-moj: When I ran the api in dev, only the note was added.  The 'created by' part was empty, see second contact for D001024 in dev.

achimber-moj: I assume 'notes += outcome.notes' will work?

marcus-bcl: I believe so, you might need to stick a new line in there as well but I'm not sure.

achimber-moj: going to deploy my branch to dev with an update to see if the notes are created successfully

PR #4455, opened 2024-11-21
APS-1482 - Refactor GET case-detail endpoint to include offence categories

gregkhawkins:
**Background**
- The `Community API`'s endpoint: `GET /secure/offenders/crn/{crn}/convictions` includes offence categories in the response
- We need these in the `Approved Premises` API to build offence descriptions

**PR Includes**
- For the `GET /probation-cases/{crn}/details` endpoint, retrieve the the extra offence categories from the DB and surface on the API response

### DIFF
--- a/projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseDetail.kt
+++ b/projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseDetail.kt
@@ -62,6 +62,8 @@ data class Offence(
     val id: String,
     val code: String,
     val description: String,
+    val mainCategoryDescription: String,
+    val subCategoryDescription: String,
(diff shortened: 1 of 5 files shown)


projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseDetail.kt
```diff
@@ -62,6 +62,8 @@ data class Offence(
     val id: String,
     val code: String,
     val description: String,
+    val mainCategoryDescription: String,
+    val subCategoryDescription: String,
```
gregkhawkins, line 66: when we get these back from community api both fields are non-nullable so have followed suit here as my understanding is that it is the same underlying database. Should I do anything else to 100% confirm this or all good?

marcus-bcl (approved):

PR #4490, opened 2024-12-12
PI-2670 - Common-platform-and-delius - Feature flag to toggle data creation 

joseph-bcl:


projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -54,19 +54,44 @@ class Handler(
...
+                eventName,
+                mapOf(
                     "hearingId" to notification.message.hearing.id,
                     "CRN" to savedEntities.person.crn,
                     "personId" to savedEntities.person.id.toString(),
+                    "firstName" to savedEntities.person.forename,
+                    "surname" to savedEntities.person.surname,
+                    "dob" to savedEntities.person.dateOfBirth.toString(),
+                    "genderCode" to savedEntities.person.gender.description,
+                    "pnc" to savedEntities.person.pncNumber.toString(),
```
marcus-bcl, line 77: We should be careful not to log PII to application insights, so probably best to stick with the identifiers in here (e.g. CRN / hearing / defendant id).

I know this makes the "dry run" events a bit useless, but we can at least get an idea of how many we would create.  And maybe we could make a call to the prepare a case system to get more details.

projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonService.kt
```diff
@@ -27,113 +28,138 @@ class PersonService(
...
-                require(age > 10) {
-                    "Date of birth would indicate person is under ten years old: $it"
-                }
+        // Under 10 years old validation
+        dateOfBirth.let {
+            val age = Period.between(it, LocalDate.now()).years
+            require(age > 10) {
+                "Date of birth would indicate person is under ten years old: $it"
             }
+        }
```
marcus-bcl, line 57: Would it simplify things to move this out into the Handler, or into a "validate" method?

That way you could perform this validation regardless of the feature flag, and only call `insertPerson` if the flag is enabled - e.g. the Handler would become something like:

```kotlin
// this could potentially be moved to a validate message:
val dateOfBirth = requireNotNull(defendant.personDefendant?.personDetails?.dateOfBirth) { "Date of birth not found in message" }
val age = Period.between(dateOfBirth, LocalDate.now()).years
require(age > 10) { "Date of birth would indicate person is under ten years old: $it" }

// only need to check the flag in one place:
if (featureFlags.enabled("common-platform-record-creation-toggle")) {
    val (person, address) = personService.insertPerson(defendant, notification.message.hearing.courtCentre.code)
    notifier.caseCreated(person)
    address?.let { notifier.addressCreated(it) }
    telemetryService.trackEvent("PersonCreated", mapOf("hearingId" to ..., "defendantId" to ..., "crn" to ..., etc))
} else {
    telemetryService.trackEvent("SimulatedPersonCreated", mapOf("hearingId" to ..., "defendantId" to ...))
}
```

I'm possibly missing some important detail though, so please let me know if I am! 😄

projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -54,19 +54,44 @@ class Handler(
             }
```
marcus-bcl, line 54: I know we haven't discussed this previously, but probably worth adding an event here too when we find a match - so we can see how many we get that already exist in Delius.

projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -50,50 +52,55 @@ class Handler(
             val matchedPersonResponse = probationSearchClient.match(matchRequest)
 
             if (matchedPersonResponse.matches.isNotEmpty()) {
+                telemetryService.trackEvent(
+                    "ProbationSearchMatchDetected",
+                    mapOf(
+                        "hearingId" to notification.message.hearing.id,
+                        "defendantId" to defendant.id,
+                    )
```
marcus-bcl, line 60: Tiny extra thing - is it worth adding the CRNs when here when we find a match?


e.g.
```suggestion
                    mapOf(
                        "hearingId" to notification.message.hearing.id,
                        "defendantId" to defendant.id,
                        "crns" to matchedPersonResponse.matches.map { it.otherIds.crn }.joinToString(", ")
                    )
```

marcus-bcl (dismissed):

PR #4498, opened 2024-12-17
MAN-113: Return MAPPA and OPD in risks

pmcphee77:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/RiskService.kt
```diff
@@ -33,14 +34,44 @@ class RiskService(
     fun getPersonRiskFlags(crn: String): PersonRiskFlags {
         val summary = personRepository.getSummary(crn)
         val riskFlags = riskFlagRepository.findByPersonId(summary.id)
+        val opd = nsiRepository.findByPersonIdAndTypeCode(summary.id, NsiType.Code.OPD_COMMUNITY_PATHWAY.value)
```
marcus-bcl, line 37: Do we need to check the NSI active flag here?

pmcphee77: yes, changed

PR #4589, opened 2025-01-27
Man 340 requirements notes truncated

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Requirement.kt
```diff
@@ -7,12 +7,13 @@ import org.hibernate.type.NumericBooleanConverter
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
 import uk.gov.justice.digital.hmpps.api.model.overview.Rar
+import uk.gov.justice.digital.hmpps.integrations.delius.referencedata.entity.ReferenceData
 import java.time.LocalDate
 
 @Immutable
 @Entity
 @Table(name = "rqmnt")
-@SQLRestriction("soft_deleted = 0 and active_flag = 1")
```
pmcphee77, line 15: Did you mean to remove this? Just checking, just means inactive records will be pulled back as part of non native queries

achimber-moj: for requirements, inactive records are shown.

pmcphee77 (approved):

PR #4594, opened 2025-01-29
Correct the assessment summary event type in the docs

jsimonsmoj:
Correct the name of the assessment summary and Delius queue name and event type filter.

### DIFF
--- a/projects/assessment-summary-and-delius/README.md
+++ b/projects/assessment-summary-and-delius/README.md
@@ -29,9 +29,9 @@ Assessment summary relies on the OASys ORDS API for retrieving the assessment su
 ## Assessment summary produced workflow
 The assessment summary is recorded in Delius. The risks are evaluated and stored as registrations.
 
-| Business Event        | Message Event Type / Filter  |
-| --------------------- | ---------------------------- |
-| Assessment submitted  | assessment.summary.produced  |
+| Business Event        | Message Event Type / Filter       |
+| --------------------- | --------------------------------- |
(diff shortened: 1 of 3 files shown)


projects/assessment-summary-and-delius/README.md
```diff
@@ -29,9 +29,9 @@ Assessment summary relies on the OASys ORDS API for retrieving the assessment su
 ## Assessment summary produced workflow
 The assessment summary is recorded in Delius. The risks are evaluated and stored as registrations.
 
-| Business Event        | Message Event Type / Filter  |
-| --------------------- | ---------------------------- |
-| Assessment submitted  | assessment.summary.produced  |
+| Business Event        | Message Event Type / Filter       |
+| --------------------- | --------------------------------- |
+| Assessment submitted  | risk-assessment.scores.determined |
```
marcus-bcl, line 34: Not sure if it's worth mentioning that we do an additional filter for `"assessment.summary.produced"`, here: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/467ffb18c01d837fa9357c05f172bbf4cca1412a/projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt#L31

Unfortunately due to the way OASys generates their SNS notifications, the `MessageAttributes.eventType` is always set to `"risk-assessment.scores.determined"`, but `Message.eventType` is set to a more detailed event type that we can filter on in the application layer.

marcus-bcl (approved):

marcus-bcl (approved):

PR #4599, opened 2025-01-29
MAN-172: Added personal contact details update

pmcphee77:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonalDetailsService.kt
```diff
@@ -26,9 +32,87 @@ class PersonalDetailsService(
...
+
+        val person = personRepository.getPerson(crn)
+        val mainAddress = addressRepository.findByPersonId(person.id)
+            .filter { it.endDate == null }
+            .firstOrNull { it.status.code == "M" }
+
+        person.telephoneNumber = request.phoneNumber
+        person.mobileNumber = request.mobileNumber
+        person.emailAddress = request.emailAddress
+        val updated = toMainAddress(person.id, mainAddress, addressType, request, startDate)
```
marcus-bcl, line 67: Does this update the address in-place?  I thought Delius would end-date the previous address and create a new one.

marcus-bcl: Also - do we need to create audit records and send domain events out for the changes?

It's probably worth looking through the Delius code to make sure we're not missing any validation or extra processing for address and personal details updates.

marcus-bcl (commented):
You'll also need to add any tables that you need write access to in the `deploy/database/access.yml` file.

achimber-moj (approved):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonalDetailsService.kt
pmcphee77: Before UI
Pasted Graphic
Before DB
Pasted Graphic 1
Updating Street Name from Paddock Road to Paddock Street
Pasted Graphic 2
Run query again to get the original address record with old street name (nothing returned)
Pasted Graphic 3
Now look for updated address using Paddock Street
Pasted Graphic 4
Same record has been updated in place

pmcphee77:
> You'll also need to add any tables that you need write access to in the `deploy/database/access.yml` file.

Fixed

marcus-bcl (approved):

PR #4602, opened 2025-01-30
PI-2752

anthony-britton-moj:


projects/prison-case-notes-to-probation/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/model/DeliusCaseNote.kt
```diff
@@ -42,16 +42,20 @@ data class CaseNoteBody(
...
             isResettlementPassport() -> BcstPathway.from(content)?.let { "BCST case note for $it" }
             caseNoteType.code == CaseNoteType.DEFAULT_CODE -> "NOMIS Case Note - $type - $subType"
             else -> null
         }
     }
 
+    private fun String.truncatedDescription(): String {
+        val bytes = this.toByteArray()
+        return if (bytes.size > 200) {
+            String(bytes.slice(0..197).toByteArray()) + " ~"
```
marcus-bcl, line 55: Is there a chance of us corrupting the end of the text here, if we remove one byte from a multi-byte character?  Not sure how best to deal with that...

projects/prison-case-notes-to-probation/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/model/DeliusCaseNote.kt
```diff
@@ -42,16 +42,20 @@ data class CaseNoteBody(
...
+        } else this
```
marcus-bcl, line 56: Bit fiddly, but maybe something like this?
```suggestion
    private fun String.truncatedDescription(): String {
      var count = 0
      return takeWhile { 
        count += it.toString().encodeToByteArray().size
        count <= 200
      }
    }
```

marcus-bcl (approved):

PR #4618, opened 2025-02-04
MAN-2787 - use disposal id to calculate rar days

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/RequirementService.kt
```diff
@@ -59,14 +59,18 @@ class RequirementService(
...
     }
 
-    fun getRarDescription(id: Long, eventNumber: String): String? {
-        val sumRarRequirementLength = requirementRepository.getRequirements(id, eventNumber)
-            .filter { it.mainCategory!!.code == "F" }
-            .sumOf { it.length!! }
+    fun getRarDescription(eventId: Long, eventNumber: String, disposalId: Long): String? {
+        val rarCode = "F"
+        val rarRequirements = requirementRepository.getRequirements(eventId, eventNumber)
+            .filter { it.mainCategory!!.code == rarCode }
```
marcus-bcl, line 65: Maybe worth moving this filter into the query?

marcus-bcl (approved):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/RequirementService.kt
achimber-moj: unfiltered query is used for another service, so decided to not repeat the query with another and clause

PR #4650, opened 2025-02-11
Pi 2803 update get contacts api

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/sentence/entity/OffenderManager.kt
```diff
@@ -36,7 +37,23 @@ class OffenderManager(
     @JoinColumn(name = "allocation_staff_id")
     val staff: Staff,
 
-    val endDate: LocalDate?,
+    val allocationDate: LocalDate,
+
+    val endDate: LocalDate? = null,
+
+    @OneToMany
```
marcus-bcl, line 44: Would it be more efficient to make this a one-to-one mapping, and add the endDate filter as a `@Where` clause?

achimber-moj: done

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/sentence/entity/OffenderManager.kt
```diff
@@ -95,6 +113,29 @@ class StaffUser(
...
+    @Id
+    @Column(name = "responsible_officer_id", nullable = false)
+    val id: Long = 0,
+
+    @Column(name = "offender_id")
+    val personId: Long,
+
+    @OneToOne
+    @JoinColumn(name = "prison_offender_manager_id")
+    var prisonManager: PrisonManager? = null,
```
marcus-bcl, line 129: Can this be a val?

marcus-bcl (approved):

PR #4674, opened 2025-02-17
PI-2836 add application origin to contact description

anthony-britton-moj:


marcus-bcl (approved):
Seems weird to output the word null in the description, rather than just omit it?

PR #4707, opened 2025-02-24
Bump Spring Boot to 3.4.3

marcus-bcl:
Also bump Flipt client to 1.1.2 and Springdoc to 2.8.5


marcus-bcl:
@dependabot show org.springframework.boot ignore conditions

PR #4713, opened 2025-02-24
Pi 2864 inactive events incorrectly returned

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Event.kt
```diff
@@ -59,7 +59,17 @@ class Event(
...
+) {
+    fun isInactiveEvent(): Boolean {
+        if (disposal != null && disposal.active.not())
+            return true
+
+        if (active.not())
+            return true
+
+        return false
+    }
```
marcus-bcl, line 71: It might be me but this is quite hard to get my head around, probably just because it's a double negative.  Maybe we could simplify it?

E.g. to

```suggestion
    fun isActiveEvent() = active && (disposal == null || disposal.active)
```

or 

```suggestion
    fun isInactiveEvent() = !active || (disposal?.active == false)
```

marcus-bcl (approved):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Event.kt
achimber-moj: i did have two methods, but the logic i needed into one, as i found them easier to read

PR #4724, opened 2025-02-26
PI-2836 updated to always have application origin and renamed field

anthony-britton-moj:


projects/cas2-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/client/approvedpremises/model/EventDetails.kt
```diff
@@ -13,29 +13,28 @@ data class ApplicationSubmitted(
...
     PrisonBail("Prison Bail"),
     CourtBail("Court Bail"),
-    HomeDetentionCurfew("Home Detention Curfew"),
-    Null("Null");
+    HomeDetentionCurfew("Home Detention Curfew");
 
     companion object {
         fun from(value: String?): ApplicationOrigin =
-            ApplicationOrigin.entries.firstOrNull { it.name.lowercase() == value?.lowercase() } ?: Null
+            ApplicationOrigin.entries.first { it.name.lowercase() == value?.lowercase() }
```
marcus-bcl, line 38: Will this throw an error if they send null?  Is that intentional?

marcus-bcl (approved):

PR #4767, opened 2025-03-07
feat: adds users for CAS2v2 (bail)

tobybatchmoj:


projects/approved-premises-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/generator/StaffGenerator.kt
```diff
@@ -29,6 +35,8 @@ object StaffGenerator {
 
     val DEFAULT_STAFF_USER = generateStaffUser("john-smith", DEFAULT_STAFF)
     val JIM_SNOW_USER = generateStaffUser("JIMSNOWLDAP", JIM_SNOW)
+    val CAS2V2_DELIUS_CB_USER = generateStaffUser("CAS2V2_DELIUS_CB", CAS2V2_DELIUS_CB)
+    val CAS2V2_DELIUS_PB_USER = generateStaffUser("CAS2V2_DELIUS_PB", CAS2V2_DELIUS_PB)
```
davidatkinsuk, line 39: I think `DELIUS` is implicit as we're in the `ap-and-delius` project, so probably don't need it in the constant name here (Same goes for other constants)

davidatkinsuk: Actually scrub that, i guess you want it in the username to differentiate between non-delius users

davidatkinsuk:
Looks good to me. You'll need a review from a member of the probation-integration team. I'd suggest posting it to #topic-pi-cas123

projects/approved-premises-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/generator/StaffGenerator.kt
tobybatchmoj: I'd like to keep these the same as they now match the user names across the other two repos that refer to these users.

PR #4837, opened 2025-03-21
PI-2831 - Fetch offence priority data from S3 and use it to determine main offence.

jamesmacbeth-unilink:


projects/common-platform-and-delius/build.gradle.kts
```diff
@@ -16,8 +16,10 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+    implementation("com.fasterxml.jackson.dataformat:jackson-dataformat-csv")
     implementation("org.opensearch.client:spring-data-opensearch-starter:1.6.1")
     implementation("org.opensearch.client:opensearch-java:2.20.0")
+    implementation("io.awspring.cloud:spring-cloud-aws-starter-s3:3.3.0")
```
marcus-bcl, line 22: Could we move these three dependency versions in to the versionCatalogs section in settings.gradle.kts?

projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/OffenceService.kt
```diff
@@ -0,0 +1,46 @@
...
+            .with(schema)
+            .readValues<OffencePriority>(s3Client.getObject(request))
+
+        val priorityMap = mutableMapOf<String, Int>()
+
+        reader.forEach { record ->
+            priorityMap[record.hoOffenceCode] = record.priority
+        }
+
+        return priorityMap
```
marcus-bcl, line 44: You could simplify this a little using Kotlin's [associateBy](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.sequences/associate-by.html) function.  E.g.
```suggestion
        return CsvMapper().readerFor(OffencePriority::class.java)
            .with(CsvSchema.emptySchema().withHeader())
            .readValues<OffencePriority>(s3Client.getObject(request))
            .asSequence()
            .associateBy({ it.hoOffenceCode }, { it.priority })
```

projects/common-platform-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/OffenceService.kt
```diff
@@ -0,0 +1,46 @@
...
+    private val s3Client: S3Client,
+    @Value("\${s3.bucket.name}") private val bucketName: String
+) {
+    val priorityMap: Map<String, Int> by lazy { readOffencePrioritiesFromS3() }
+
+    fun getOffenceHomeOfficeCodeByCJACode(cjaCode: String): String =
+        detailedOffenceRepository.findByCode(cjaCode)?.homeOfficeCode?.replace(
+            "/",
+            ""
+        ) ?: throw IllegalArgumentException("No Home Office code found for CJA code $cjaCode")
```
marcus-bcl, line 24: Minor formatting thing.  IntelliJ's formatter sometimes wraps things a bit weird - would probably be clearer to wrap at the `?:`
```suggestion
    fun getOffenceHomeOfficeCodeByCJACode(cjaCode: String): String =
        detailedOffenceRepository.findByCode(cjaCode)?.homeOfficeCode?.replace("/", "") 
            ?: throw IllegalArgumentException("No Home Office code found for CJA code $cjaCode")
```

The formatter should leave it alone then.

marcus-bcl (approved):

PR #4857, opened 2025-03-26
CBA-412/fix-seed-data-for-local-to-match-dev

tobybatchmoj:
feat: adds a user to match the NOMS/CRN of an excluded offender in dev env


davidatkinsuk (approved):
LGTM

PR #4908, opened 2025-04-09
PI-2976 create new search endpoints for SOC

anthony-britton-moj:


projects/soc-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/SearchRequest.kt
```diff
@@ -0,0 +1,39 @@
...
+    val pncNumber: String? = null,
+    val crn: String? = null,
+    val nomsNumber: String? = null,
+)
+
+@Target(AnnotationTarget.CLASS)
+@Retention(AnnotationRetention.RUNTIME)
+@Constraint(validatedBy = [SearchRequestValidator::class])
+annotation class ValidSearchRequest(
+    val message: String = "At least one decision field or at least one action must be provided",
```
marcus-bcl, line 23: What is decision / action here?

projects/soc-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProbationCaseSearch.kt
```diff
@@ -0,0 +1,48 @@
...
+    fun find(request: SearchRequest): ProbationCases =
+        ProbationCases(personRepository.findAll(request.asSpecification()).map { it.toProbationCase() })
+
+    fun crns(crns: Set<String>): ProbationCases =
+        ProbationCases(personRepository.findByCrnIn(crns).map { it.toProbationCase() })
+}
+
+private fun SearchRequest.asSpecification(): Specification<DetailPerson> = listOfNotNull(
+    firstName?.let { matchesForename(it) },
+    surname?.let { matchesSurname(it) },
```
marcus-bcl, line 20: Do we need to include aliases here?  Or does the search endpoint they're currently using not do that?

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #4910, opened 2025-04-09
PI-2974: Move probation match to db

pmcphee77:
This PR provides 2 feature flags
**prison-identifiers-use-db-search**
When set to true, this will replace the call to the probation offender API match with a set of DB queries that return the same model.
**prison-identifiers-compare-with-db-search**
When set to true and _prison-identifiers-use-db-search_ is set to false, this will continue to use the probation offender API match to match the offender, but will also perform the equivalent DB match for comparison. If there are differences in the match result, a _MatchDifferenceFound_ app insights event is raised.

### DIFF
--- a/projects/prison-identifier-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationSearchRepository.kt
+++ b/projects/prison-identifier-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationSearchRepository.kt
@@ -203,16 +203,6 @@ interface ProbationSearchRepository : JpaRepository<Person, Long> {
     ): List<Person>
 }
 
-/*
-ProbationMatchRequest(
-firstName=TERRY,
-surname=BEIER-HUEL,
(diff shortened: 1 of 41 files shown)


projects/prison-identifier-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationSearchRepository.kt
```diff
@@ -0,0 +1,266 @@
...
+/*
+ProbationMatchRequest(
+firstName=TERRY,
+surname=BEIER-HUEL,
+ dateOfBirth=1973-04-29,
+ nomsNumber=A1036EA,
+ activeSentence=true,
+ pncNumber=1973/636766X,
+  croNumber=null)
+ */
```
marcus-bcl, line 215: Is this comment supposed to be here?

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #4960, opened 2025-04-25
PI-2974: Fixed cro and pnc DB queries

pmcphee77:


projects/prison-identifier-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationSearchRepository.kt
```diff
@@ -108,12 +108,14 @@ interface ProbationSearchRepository : JpaRepository<Person, Long> {
...
-                    or p.dateOfBirth = :dateOfBirth
-                    or a.dateOfBirth = :dateOfBirth)   
+                (:surname is not null and (
+                    lower(trim(p.surname)) = lower(trim(:surname)) 
+                    or lower(trim(a.surname)) = lower(trim(:surname))
+                ))
+                or (:dateOfBirth is not null and ( 
+                    p.dateOfBirth = :dateOfBirth
+                    or a.dateOfBirth = :dateOfBirth)
+                )
```
marcus-bcl, line 118: I might be misreading this, but do you just need to change it from `is not null` to `is null` - so skip the check if a surname filter isn't provided?
```suggestion
                (:surname is null 
                    or lower(trim(p.surname)) = lower(trim(:surname)) 
                    or lower(trim(a.surname)) = lower(trim(:surname)))
                or (:dateOfBirth is null 
                    or p.dateOfBirth = :dateOfBirth
                    or a.dateOfBirth = :dateOfBirth)  
```

pmcphee77: I was wrestling with this for a while and was originally doing as you suggested but it was causing issues. Imagine if they did not provide a surname, but they did provide a date of birth. It would resolve to true in the surname check and would then completely ignore the date of birth check. So checking first that its not null before doing the or on the alias and offender, then it will still do the DOB check if the surname is null

marcus-bcl (approved):

PR #4981, opened 2025-05-01
MAN-563: Add contactOutcomeFlag to search index

pmcphee77:


projects/person-search-index-from-delius/container/pipelines/contact/statement.sql
```diff
@@ -30,6 +30,7 @@ from (with page as (select * from contact where :contact_id = 0
                                                  then 'Y' else 'N' end,
                      'nationalStandard' value r_contact_type.national_standards_contact,
                      'systemGenerated' value decode(r_contact_type.sgc_flag, 1, 'Y', 0, 'N', null),
+                     'contactOutcomeFlag' value decode(r_contact_type.contact_outcome_flag, 1, 'Y', 0, 'N', null),
```
marcus-bcl, line 33: Could we drop the contact prefix on the field name like we have for the other fields e.g. `'outcomeFlag'` or `'outcomeRequiredFlag'` might make more sense.

Also, the value in the database can by `Y`, `N` or `B`, but you're mapping `B` to `null` - is that intentional?

marcus-bcl (approved):

projects/person-search-index-from-delius/container/pipelines/contact/statement.sql
pmcphee77: Changed the name, we are only interested in where the outcome is required

PR #4995, opened 2025-05-06
Pi 3041 incorrect entity type being passed to alfresco

achimber-moj:


marcus-bcl (approved):
Great, nice and simple!

PR #5050, opened 2025-05-19
PI-3012 address memory issues

anthony-britton-moj:


templates/helm-defaults.yml
```diff
@@ -5,7 +5,10 @@ generic-service:
   resources:
     requests:
       cpu: 128m
-      memory: 512Mi
+      memory: 1024Mi
+    limits:
+      cpu: 1024m
```
marcus-bcl, line 10: Not sure if it was intentional, but this is slightly more than 1 CPU (1024 millicores == 1.024 cores).  

If you were after a single core, then you can just set it to "1" (or 1000 millicores).
```suggestion
      cpu: "1" # or 1000m
```

This is currently set as the default here: https://github.com/ministryofjustice/cloud-platform-environments/blob/6d477a1f988ac2b318b1e3a92c9cf0b1ebb84fb1/namespaces/live.cloud-platform.service.justice.gov.uk/hmpps-probation-integration-services-prod/02-limitrange.yaml#L9

templates/helm-defaults.yml
```diff
@@ -5,7 +5,10 @@ generic-service:
   resources:
     requests:
       cpu: 128m
-      memory: 512Mi
+      memory: 1024Mi
```
marcus-bcl, line 8: Is this a bit high, given we're averaging under 700 in prod?
```
$ kubectl -n hmpps-probation-integration-services-prod top pod | awk '{sum+=$3} END {print "Avg: " sum/NR}'
Avg: 695.882
```

Maybe 768Mi would be more appropriate?

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #5066, opened 2025-05-23
PI-3085: refactor logic to reduce number of db calls

achimber-moj:


projects/breach-notice-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/Contact.kt
```diff
@@ -71,6 +71,10 @@ class Contact(
     @Id
     @Column(name = "contact_id")
     val id: Long,
+
+    @OneToMany
+    @JoinColumn("contact_id")
+    val unpaidWorkAppointments: List<UpwAppointment>? = emptyList(),
```
marcus-bcl, line 77: Should this be a `mappedBy`, as the join column is on the other entity?
```suggestion
    @OneToMany(mappedBy = "contact")
    val unpaidWorkAppointments: List<UpwAppointment> = emptyList(),
```

projects/breach-notice-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/UpwAppointment.kt
```diff
@@ -15,12 +14,9 @@ class UpwAppointment(
     @Column(name = "upw_appointment_id")
     val id: Long,
 
+    @Column("contact_id")
     val contactId: Long,
```
marcus-bcl, line 18: Think the JoinColumn should be here:
```suggestion
    @ManyToOne
    @JoinColumn("contact_id")
    val contact: Contact,
```

projects/breach-notice-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/WarningService.kt
```diff
@@ -42,19 +41,14 @@ class WarningService(
...
-
         return WarningDetails(
             breachReasons = breachReasons.codedDescriptions(),
             enforceableContacts = enforceableContacts.map(Contact::toEnforceableContact),
-            unpaidWorkRequirements = unpaidWorkRequirements
+            unpaidWorkRequirements = enforceableContacts.filter { it.unpaidWorkAppointments?.isNotEmpty() == true }
+                .let {
+                    requirementRepository.getUnpaidWorkRequirementsByDisposal(disposal.id)
+                        .map { it.toModel() }
+                }
```
marcus-bcl, line 51: Not sure if this is doing what you want, the `.let` will always run because `.filter` always returns a non-null list.

I think this would work:
```suggestion
            unpaidWorkRequirements = if (enforceableContacts.any { it.unpaidWorkAppointments.isNotEmpty() == true }) {
                requirementRepository.getUnpaidWorkRequirementsByDisposal(disposal.id)
                    .map { it.toModel() }
            } else emptyList()
```

achimber-moj: i am still making changes, will raise a pr when i am done

achimber-moj: was changing the code to use any

projects/breach-notice-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/UpwAppointment.kt
achimber-moj: done

projects/breach-notice-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/Contact.kt
achimber-moj: done

marcus-bcl (approved):

PR #5073, opened 2025-05-28
PI-2991 migrate court case from pos

anthony-britton-moj:


projects/court-case-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PncNumber.kt
```diff
@@ -0,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.service
+
+import java.time.LocalDate.now
+
+data class PncNumber(val year: Int, val serial: Int, val checkSum: Char) {
+    fun matchValue() = "${fullYear()}/${String.format("%07d", serial)}" + checkSum.uppercase()
+    private fun fullYear() = when {
+        year > 99 -> year
+        year % 100 <= now().year % 100 -> "20${formattedYear()}"
```
marcus-bcl, line 9: Minor thing - this modulo isn't required, is it?  We already know `year` is <100 at this point.
```suggestion
        year <= now().year % 100 -> "20${formattedYear()}"
```

marcus-bcl (dismissed):

PR #5194, opened 2025-06-27
Pi 3192 add default provider and team to payload

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/user/entity/User.kt
```diff
@@ -195,6 +195,35 @@ interface TeamRepository : JpaRepository<Team, Long> {
...
+            SELECT t
+            FROM Team t
+            WHERE t.provider.code = :providerCode
+            AND t.id = :id
+            AND (t.endDate IS NULL OR t.endDate > CURRENT_DATE)
+            AND t.startDate <= CURRENT_DATE
+            ORDER BY UPPER(t.description) 
+        """
+    )
+    fun findByProviderCodeAndTeamId(providerCode: String, id: Long): Team?
```
marcus-bcl, line 210: The team id is unique, do we need to filter on provider code as well?  Also the start/end date filtering on the team is duplicated in each query - is it worth moving that to a `@SqlRestriction` on the entity?  Then you can just do:
```suggestion
    fun findTeamById(id: Long): Team?
```

achimber-moj: Done.  Reason I put provider code in as well as to make sure the team was part of the provider.  However,  I guess the preferences would never have a team associated with another providers

marcus-bcl (approved):

PR #5297, opened 2025-07-29
Pi 3313 appointment created by

achimber-moj:


achimber-moj:
> Is there a reason we can't just add the `@CreatedBy` annotation to the `Appointment.createdByUserId` field?

using the annotation results in the field being updated with the service user id , e.g id for username ManageSupervisionAndDelius

marcus-bcl (approved):

PR #5302, opened 2025-07-31
PI-3275 Switch to soundex for name matching

anthony-britton-moj:


projects/prison-identifier-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/ProbationSearchRepository.kt
```diff
@@ -27,9 +30,9 @@ interface ProbationSearchRepository : JpaRepository<Person, Long> {
                 and
                 (:activeSentence != true or (:activeSentence = true and p.currentDisposal = true ))
                 and
-                (:forename is null or (:forename is not null and lower(trim(p.forename)) = lower(trim(:forename))))
+                (:forename is null or p.firstNameSoundex = :forename)
```
marcus-bcl, line 33: I think you could also use the JPA function to call soundex, to save the extra calls to get the soundex up-front:
```suggestion
                (:forename is null or p.firstNameSoundex = function('SOUNDEX', :forename))
```

marcus-bcl (approved):

PR #5311, opened 2025-08-01
PI-3332: get contacts in descending order by default

achimber-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/client/ProbationSearchClient.kt
```diff
@@ -11,7 +11,8 @@ interface ProbationSearchClient {
     fun contactSearch(
         @RequestBody body: ActivitySearchRequest,
         @RequestParam page: Int = 0,
-        @RequestParam size: Int = 10
+        @RequestParam size: Int = 10,
+        @RequestParam sort: String
```
harrytmthy, line 15: Would it make sense to make add a default value for `sort`?

PR #5399, opened 2025-08-20
PI-3356 move to correct offender on update

anthony-britton-moj:


projects/prison-case-notes-to-probation/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/service/DeliusService.kt
```diff
@@ -44,9 +44,12 @@ class DeliusService(
...
 
             ActiveAlert, InactiveAlert -> caseNoteRepository.findByExternalReference(caseNote.urn)
-        }?.takeIf {
-            //get a case note and need to ensure the noms id matches the delius noms id
-            it.offender.nomsId == caseNote.header.nomisId
+        }?.let {
+            if (it.offender.nomsId == caseNote.header.nomisId) it
+            else {
+                caseNoteRepository.save(it.dataCleanseContact())
+                caseNoteRepository.save(caseNote.newEntity().copy(id = it.id, version = it.version))
```
marcus-bcl, line 51: Should this be updating the offenderId, rather than the contact id?
```suggestion
                caseNoteRepository.save(caseNote.newEntity().copy(offenderId = it.offenderId))
```

Also, I wonder if it would be possible to output a single data cleanse contact per domain event we've acted on, rather than per contact, if that makes sense?  i.e. if we receive a person identifier message which results in us moving 100 contacts, then we shouldn't create 100 data cleanse contacts on the original person.

marcus-bcl (approved):

PR #5400, opened 2025-08-20
PI-3368 add lau detail

anthony-britton-moj:


projects/soc-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProbationAreaService.kt
```diff
@@ -35,7 +35,7 @@ class ProbationAreaService(
             .mapValues { (_, v) -> v.mergeOverlapping().sortedBy { it.startDate } }
             .toMap()
 
-    private fun List<ManagerHistory>.mergeOverlapping() = groupBy { it.probationArea.code }.values
+    private fun List<ManagerHistory>.mergeOverlapping() = groupBy { it.team.district.borough.probationArea.code }.values
```
marcus-bcl, line 38: Is this still merging by probation area?  I think we need to break it down by LAU instead now.

So if we had a history that looks like:
* LAU1, PDU1, PA1
* LAU1, PDU1, PA1
* LAU2, PDU1, PA1
* LAU3, PDU2, PA2

then we should return three history items.  But I think currently the code will return only two?

PR #5411, opened 2025-08-26
PI-3364 listen for e supervision received and expired events

anthony-britton-moj:


projects/esupervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/ESupervisionService.kt
```diff
@@ -0,0 +1,47 @@
...
+        com.person,
+        event,
+        contactTypeRepository.getByCode(ContactType.E_SUPERVISION_CHECK_IN),
+        occurredAt.toLocalDate(),
+        occurredAt,
+        com.provider,
+        com.team,
+        com.staff,
+        description(),
+        "For more information please visit ${checkInUrl()}",
```
jsimonsmoj, line 43: The required wording for the notes has been updated to `Please follow this link to review the check-in in the E-supervision portal: ${checkInUrl()}`

These is also a requirement for the notes: "Automatically generated by the E-Supervision services at: [date and time]"

jsimonsmoj (commented):
The contact notes need to be updated to align with a recent change in requirements

jsimonsmoj (approved):

PR #5476, opened 2025-09-18
PI-3431 - Created Treatment page endpoint to get documents linked to contacts

jamesmacbeth-unilink:
…provided contact ids


projects/suicide-risk-form-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/DocumentService.kt
```diff
@@ -53,6 +56,27 @@ class DocumentService(
         alfrescoUploadClient.delete(document.alfrescoId)
     }
 
+    fun listDocumentsForContacts(contactIds: List<Long>): ContactDocumentResponse {
+        val contactDocs = contactIds.map {
+            val documents = documentRepository.findByTableNameAndPrimaryKeyId("CONTACT", it)
```
marcus-bcl, line 61: Would it be better to move the query up here, so rather than executing it N times, we just execute it once?

E.g.

```suggestion
        val contactDocs = documentRepository.findByTableNameAndPrimaryKeyIdIn("CONTACT", contactIds)
            .associateBy { it.primaryKeyId }
            .map {
              ...
            }
```

marcus-bcl (approved):

PR #5513, opened 2025-09-30
PI-3475 switch to hmmps feature toggles

anthony-britton-moj:


README.md
marcus-bcl, line 211: Does this one need updating too?


```suggestion
Feature flags are managed in the [Flipt dashboard](https://feature-toggles.hmpps.service.justice.gov.uk).
```

PR #5647, opened 2025-10-20
PI-3533 - Added Community Payback Teams endpoint

jamesmacbeth-unilink:


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/TeamsResponse.kt
```diff
@@ -0,0 +1,17 @@
...
+import uk.gov.justice.digital.hmpps.integrations.delius.Team
+
+data class TeamsResponse(
+    val teams: List<TeamDTO>
+)
+
+data class TeamDTO(
+    val code: String,
+    val description: String
+)
```
marcus-bcl: Is it worth making this a generic `CodeDescription`-type class? I expect there'll be a few other response objects in this format from the other endpoints.

PR #5664, opened 2025-10-27
PI-3509 - Added Get Supervisors endpoint for community payback and delius

jamesmacbeth-unilink:


marcus-bcl:
@jamesmacbeth-unilink it might be worth moving the entities into a `entity` package - so they're excluded from the coverage report.  That should get the coverage back up in Sonar: https://sonarcloud.io/component_measures?id=ministryofjustice_hmpps-probation-integration-services&branch=PI-3509&metric=new_coverage&view=list

PR #5727, opened 2025-11-11
update to activity search

anthony-britton-moj:


anthony-britton-moj:
No longer required as MPOP UI now updated to use isAppointment flag and determine what should be displayed in the UI

PR #5743, opened 2025-11-14
PI-3545 - Update UPW Appointment Outcome endpoint

jamesmacbeth-unilink:


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/UnpaidWorkAppointment.kt
```diff
@@ -9,34 +9,35 @@ import org.springframework.data.annotation.CreatedDate
...
 import uk.gov.justice.digital.hmpps.model.AppointmentResponseCase
 import uk.gov.justice.digital.hmpps.model.AppointmentResponseName
+import uk.gov.justice.digital.hmpps.model.CodeDescription
 import java.time.LocalDate
 import java.time.LocalTime
 import java.time.ZonedDateTime
 
 @Entity
 @Table(name = "upw_appointment")
-@Immutable
```
anthony-britton-moj: Are we actually creating these now? I think you'll need to add the sequence used to the table if that is the case?

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/Contact.kt
```diff
@@ -3,32 +3,115 @@ package uk.gov.justice.digital.hmpps.integrations.delius.entity
...
+import java.time.LocalTime
 
 @Entity
 @Table(name = "contact")
-@Immutable
 class Contact(
     @Id
     @Column(name = "contact_id")
-    val id: Long,
+    @SequenceGenerator(name = "contact_id_seq", sequenceName = "contact_id_seq", allocationSize = 1)
```
anthony-britton-moj: You'll also need to add the tables you write to in the deploy -> database -> access.yaml file

anthony-britton-moj (changes requested):

PR #5753, opened 2025-11-18
PI-3608 updating created_datetime to match last_saved

sambailey-moj:


projects/breach-notice-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/Document.kt
```diff
@@ -34,6 +35,8 @@ class Document(
 
     var lastSaved: ZonedDateTime? = null,
 
+    var createdDatetime: LocalDateTime? = null,
```
marcus-bcl: Could we use a ZonedDateTime for this?

We made a decision a while back to use ZonedDateTime over LocalDateTime wherever possible - https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/doc/adr/0003-use-zoned-datetimes.md.  We have a custom deserializer to ensure we can handle Zoned/Local inputs interchangeably: [ZonedDateTimeDeserializer.kt](https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/871a34ba347c7df4373ec8d0417a189d5dabb0d9/libs/commons/src/main/kotlin/uk/gov/justice/digital/hmpps/datetime/ZonedDateTimeDeserializer.kt)

sambailey-moj: updated to ZonedDateTime

PR #5765, opened 2025-11-19
PI-3627 - Changed update appointment request fields to be nullable.

jamesmacbeth-unilink:


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AppointmentsService.kt
```diff
@@ -119,58 +119,55 @@ class AppointmentsService(
         )
     }
 
+    @Transactional
     fun updateAppointmentOutcome(
         projectCode: String,
         appointmentId: Long,
         appointmentOutcome: AppointmentOutcomeRequest
```
marcus-bcl: Minor thing - could we rename this parameter to `request`?  I think it would make the rest of the code a bit clearer.

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AppointmentsService.kt
```diff
@@ -119,61 +118,55 @@ class AppointmentsService(
...
+            referenceDataRepository.getBehaviour(request.behaviour.code)
+        }
 
         val staff =
-            appointmentOutcome.supervisor?.let {
-                staffRepository.findStaffByCode(appointmentOutcome.supervisor.code)
-                    ?: throw IllegalStateException("Staff with code ${appointmentOutcome.supervisor.code} not found")
+            request.supervisor.let {
+                staffRepository.getStaff(request.supervisor.code)
             }
```
marcus-bcl: You don't need the let here, now that it's non-nullable
```suggestion
        val staff = staffRepository.getStaff(request.supervisor.code)
```

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AppointmentsService.kt
```diff
@@ -249,9 +243,11 @@ class AppointmentsService(
         rowVersion = request.version.mostSignificantBits
     }
 
-    private fun Contact.update(request: AppointmentOutcomeRequest) = apply {
+    private fun Contact.update(request: AppointmentOutcomeRequest, contactOutcome: ContactOutcome?, staff: Staff) = apply {
         startTime = request.startTime
         endTime = request.endTime
+        this.staff = staff
+        this.contactOutcome = contactOutcome
         notes = notes?.let { it + System.lineSeparator() + System.lineSeparator() + request.notes }
```
marcus-bcl: Not related to this PR, but does this work if there are no existing notes?  You might be better off with something like:
```
notes = listOfNotNull(notes, request.notes).joinToString("\n\n")
```

PR #5769, opened 2025-11-21
PI-3628 - Recalculate FTC count on appointment update

jamesmacbeth-unilink:


marcus-bcl (dismissed):

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AppointmentsService.kt
```diff
@@ -254,8 +257,36 @@ class AppointmentsService(
...
             sensitive = request.sensitive
             alertActive = request.alertActive
             rowVersion = request.version.leastSignificantBits
         }
+
+    private fun Contact.updateFailureToComplyCount() {
+        if (event == null) return
+        event.ftcCount = contactRepository.countFailureToComply(event)
+
+        val ftcLimit = event.disposal?.type?.ftcLimit ?: return
```
anthony-britton-moj: is it worth doing this before counting the failure to comply if you're going to return if it's null?

PR #5777, opened 2025-11-24
PI-3636 - adding in religionHistory

sambailey-moj:


anthony-britton-moj (changes requested):

projects/core-person-record-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integration/delius/entity/ReligionHistory.kt
```diff
@@ -0,0 +1,32 @@
...
+@Immutable
+@Entity
+@Table(name = "offender_religion_history")
+class ReligionHistory(
+    @Id
+    @Column(name = "offender_religion_history_id")
+    val id: Long = 0,
+
+    @Column(name = "offender_id")
+    val offenderId: Long = 0,
```
marcus-bcl: Minor thing - but we usually rename this to something like `personId`, as the term `offender` was scrubbed from most of Delius (although it still appears in the DB tables/columns).
```suggestion
    val personId: Long = 0,
```

sambailey-moj: fixed

projects/core-person-record-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonMapping.kt
```diff
@@ -12,6 +13,7 @@ fun Person.detail(
     restrictions: LimitedAccess? = null,
     sentences: List<Sentence>,
     additionalIdentifiers: List<Identifier>,
+    religionHistory: List<ReligionHistory>? = null,
```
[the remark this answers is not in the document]
sambailey-moj: sorted

anthony-britton-moj (approved):

PR #5786, opened 2025-11-25
PI-3649 - adding in check for functional PDUs

sambailey-moj:


projects/accredited-programmes-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/staff/ProbationDeliveryUnit.kt
```diff
@@ -24,6 +22,8 @@ class ProbationDeliveryUnit(
 
     @Convert(converter = YesNoConverter::class)
     val selectable: Boolean,
+
+    val functional: Char
```
anthony-britton-moj: If you use the YesNoConverter (see selectable above) you can actually make this a boolean

marcus-bcl (approved):

PR #5791, opened 2025-11-26
PI-3632 two new endpoints for esupervision contact details

sambailey-moj:


projects/esupervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTest.kt
```diff
@@ -67,4 +80,38 @@ internal class IntegrationTest {
         assertThat(contact.isSensitive).isEqualTo(false)
         assertThat(contact.notes).isEqualTo("Check in has not been submitted on time" + System.lineSeparator() + "Review the online check in using the manage probation check ins service: https://esupervision/check-in/expired")
     }
+
+    @Test
+    fun `get contact details for a single crn`() {
+        mockMvc.perform(get("/case/${PersonContactDetailsGenerator.DEFAULT_PERSON_CONTACT_DETAILS.crn}").withToken())
+            .andExpect(status().isOk)
```
anthony-britton-moj: is it worth adding some validation that the expected properties are returned correctly? Looks like the other tests only do counts - so perhaps adding some property checks here validates that fields are correctly poplated - eg first name and surname are not switched somehow, if someone did change a constructor order in the future this would highlight the change - as it stands we have nothing that confirms the output is as expected.

projects/esupervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/ContactDetailsController.kt
```diff
@@ -0,0 +1,24 @@
...
+    @Operation(summary = "Gets contact details for a person on probation, for a case by CRN")
+    fun getContactDetails(@PathVariable crn: String) =
+        contactDetailsService.getContactDetailsForCrn(crn)
+            ?: throw ResponseStatusException(HttpStatus.NOT_FOUND, "Contact details not found")
+
+    @PreAuthorize("hasRole('PROBATION_API__ESUPERVISION__CASE_DETAIL')")
+    @PostMapping(value = ["/cases"])
+    @Operation(summary = "Gets contact details for people on probation, for multiple cases by CRN")
+    fun getContactDetailsForCases(@RequestBody crns: List<String>) =
+        contactDetailsService.getContactDetailsForCrns(crns)
```
marcus-bcl: Could we add validation here, to ensure no more than 500 CRNs are passed?

```suggestion
    fun getContactDetailsForCases(
        @RequestBody
        @Size(min = 1, max = 500, message = "Please provide between 1 and 500 crns")
        crns: List<String>
    ) = contactDetailsService.getContactDetailsForCrns(crns)
```

marcus-bcl (commented):
One small comment - otherwise looks good

projects/esupervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTest.kt
sambailey-moj: test extended now

anthony-britton-moj (approved):

PR #5792, opened 2025-11-27
PI-3655 - Set soft deleted, partitionAreaId, created/last updated info and actionTakenDate/time when creating new Enforcement.

jamesmacbeth-unilink:


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/Enforcement.kt
```diff
@@ -24,6 +30,35 @@ class Enforcement(
     @Column(name = "response_date")
     val responseDate: LocalDate? = null,
 
-    )
+    @Column(updatable = false, columnDefinition = "number")
+    @Convert(converter = NumericBooleanConverter::class)
+    val softDeleted: Boolean = false,
+
+    @Column(name = "action_taken_date")
+    val actionTakenDate: LocalDateTime = LocalDateTime.now(),
```
marcus-bcl: Is there any reason not to use a ZonedDateTime for these dates/times, like we do for the other Enforcement entities?

PR #5794, opened 2025-11-27
PI-3656 new endpoint for basic validation for pop's details

sambailey-moj:


projects/esupervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/PersonalDetails.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.model
+
+data class PersonalDetails(
+    val crn: String,
+    val name: Name,
+    val dateOfBirth: String
```
anthony-britton-moj: Is there a reason to not use a LocalDate for the dateOfBirth rather than a string?

```suggestion
    val dateOfBirth: LocalDate
```

projects/esupervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/PersonalDetails.kt
```diff
@@ -0,0 +1,7 @@
+package uk.gov.justice.digital.hmpps.model
```
anthony-britton-moj:
```suggestion
package uk.gov.justice.digital.hmpps.model

import java.time.LocalDate
```

PR #5856, opened 2025-12-09
PI-3645 Terminate requirement / licence condition on programme completion

marcus-bcl:


projects/accredited-programmes-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/sentence/component/SentenceComponent.kt
```diff
@@ -0,0 +1,28 @@
...
+    val startDate: ZonedDateTime
+    val commencementDate: ZonedDateTime?
+    var terminationDate: ZonedDateTime?
+    var terminationReason: ReferenceData?
+    var pendingTransfer: Boolean
+    val manager: Manager?
+    val pendingTransfers: List<Transfer>
+    var notes: String?
+    val type: String
+        get() = when (this) {
```
anthony-britton-moj: I'd consider making this abstract and having each SentenceComponent implement the description rather than where clause here. There are only 3 options so not necessary to change.

anthony-britton-moj (approved):

PR #5876, opened 2025-12-18
PI-3586: emit MAPPA domain events for CONTACT updates and deletes

serdaroguzhanakin:
### Summary

This change emits MAPPA domain events for CONTACT updates and deletes when the ViSOR flag is enabled.

**Rules applied:**

**CONTACT DELETE**
- ViSOR = true → MAPPA deleted domain event
- ViSOR = false → no MAPPA domain event

**OTHERWISE**
- ViSOR = true → MAPPA updated domain event
- ViSOR = false → no MAPPA domain event

**Hard delete (contact record missing)**
- If the contact record no longer exists and ViSOR cannot be resolved, no MAPPA domain event is emitted (interim decision).

Notifications and telemetry are always published regardless of ViSOR.

Unit and integration tests cover all ViSOR, soft delete, hard delete, update, and delete scenarios.


projects/offender-events-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/offender/Contact.kt
```diff
@@ -18,6 +17,14 @@ class Contact(
     @Column(name = "soft_deleted", columnDefinition = "number", nullable = false)
     @Convert(converter = NumericBooleanConverter::class)
     val softDeleted: Boolean = false,
+
+    @Column(name = "visor_exported")
+    @Convert(converter = YesNoConverter::class)
+    val visorExported: Boolean? = false
```
marcus-bcl: Apologies I missed this in our earlier chats - this should be the `visor_contact` column.
```suggestion
    @Column(name = "visor_contact")
    @Convert(converter = YesNoConverter::class)
    val visorContact: Boolean? = false
```

projects/offender-events-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTest.kt
```diff
@@ -52,6 +57,293 @@ internal class IntegrationTest @Autowired constructor(
...
+
+        offenderDeltaRepository.save(delta)
+
+        // when
+        verify(offenderDeltaService, after(500).atLeastOnce()).notify(any())
+
+        // then
+        await()
+            .atMost(5, TimeUnit.SECONDS)
+            .untilAsserted {
```
marcus-bcl: Do you need to use await here?

marcus-bcl (commented):
Looking good, thanks Serdar. I've added a few initial comments. I haven't had a proper dig through the tests yet (they're quite verbose!), but I'll keep looking.

No rush on getting these sorted, it can wait till the New Year if you want to focus on the security work you're doing.

projects/offender-events-and-delius/src/test/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/offender/OffenderDeltaServiceTest.kt
```diff
@@ -0,0 +1,782 @@
...
+            registrationRepository = registrationRepository,
+        )
+    }
+
+    @ParameterizedTest
+    @MethodSource("nonContactTables")
+    fun `when sourceTable is not CONTACT then no domain event is generated`(table: String) {
+        // given
+        val offender = Offender(
+            id = 99L,
```
marcus-bcl: Minor thing - given there are no assertions on specific IDs in this test class, you could replace all these random numbers with `id()` to make things a bit cleaner.

marcus-bcl (approved):

PR #5877, opened 2025-12-19
PI-3712 - Added support for creating appointment with type 'PRE_GROUP_ONE_TO_ONE_MEETING'

jamesmacbeth-unilink:
Update requirement/licence condition commencement_date when appointment type is PRE_GROUP_ONE_TO_ONE_MEETING. Create ECOM contact.

### DIFF
--- a/projects/accredited-programmes-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/DataLoader.kt
+++ b/projects/accredited-programmes-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/DataLoader.kt
@@ -64,6 +64,8 @@ class DataLoader(dataManager: DataManager) : BaseDataLoader(dataManager) {
         save(TestData.ENFORCEMENT_REVIEW_CONTACT_TYPE)
         save(TestData.COMPONENT_TERMINATED_CONTACT_TYPE)
         save(TestData.COMPONENT_TRANSFER_REJECTED_CONTACT_TYPE)
+        save(TestData.PRE_GROUP_ONE_TO_ONE_MEETING_CONTACT_TYPE)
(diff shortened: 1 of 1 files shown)


projects/accredited-programmes-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AppointmentService.kt
```diff
@@ -78,6 +86,39 @@ class AppointmentService(
...
+                saveComponentCommencedContact(
+                    event,
+                    requirement,
+                    licenceCondition,
+                    it,
+                    team.provider,
+                    team,
+                    staff[it.staff.code].orNotFoundBy("code", it.staff.code)
+                )
+            }
```
marcus-bcl: This `asEntities()` method was only mapping to entities before, but we're now adding the side effect of saving some entities while mapping.  It might be better to change the `map` into a `flatMap` and return a list of entities (the appointment + the component commenced contact).

Also, I'm not sure why we need to have this if statement twice?  Could we simplify things by passing `requirement ?: licenceCondition`?  Maybe we could make use of extension functions too?  E.g.

```
    private fun List<CreateAppointmentRequest>.asEntities(): List<Contact> {
            val staffMember = staff[it.staff.code].orNotFoundBy("code", it.staff.code)
            // ...
            
            if (it.type == CreateAppointmentRequest.Type.PRE_GROUP_ONE_TO_ONE_MEETING) {
                val component = requirement ?: licenceCondition
                component?.commenceComponent(event, it, team, staffMember)
            }
            
            // ...
    }
    private fun SentenceComponent.commenceComponent(
        event: Event,
        request: CreateAppointmentRequest,
        team: Team,
        staff: Staff
    ): Contact {
            val commencementDateUpdated = this.updateCommencementDate(request.date)
            if (commencementDateUpdated) {
                Contact(...
```

PR #5879, opened 2025-12-19
PI-3721 suppress CVE-2025-66293 until 2026-08-01

serdaroguzhanakin:


.trivyignore
```diff
@@ -3,3 +3,9 @@
 # Reason: Not processing any untrusted XML
 # Package: libexpat:2.6.4-r0
 CVE-2024-8176 exp:2025-08-01
```
marcus-bcl: We may as well remove the above suppression now too as it's expired?

marcus-bcl (dismissed):

PR #5913, opened 2025-12-31
PI-3507 Workforce reallocations domain event consumer

sambailey-moj:


projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/workforceallocations/AllocationDetail.kt
```diff
@@ -14,6 +14,7 @@ sealed interface AllocationDetail {
     val createdDate: ZonedDateTime
     val datasetCode: DatasetCode
     val code: String
+    val allocationReason: String?
```
marcus-bcl: Can we make this be of type AllocationReason?  Then Jackson will automatically handle the mapping for you and you can remove the `fromTextOrDefaultInitial` method.

```suggestion
    val allocationReason: AllocationReason
```

Also, Gary confirmed they won't send `null` - they'll default to `INITIAL_ALLOCATION` for us: https://mojdt.slack.com/archives/C035YK9FFK4/p1765973148040839?thread_ts=1765963573.726079&cid=C035YK9FFK4

projects/workforce-allocations-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/allocations/AllocationValidator.kt
```diff
@@ -40,9 +42,20 @@ class AllocationValidator(
...
+        val allocationType = when (allocationDetail) {
+            is AllocationDetail.PersonAllocation -> AllocationType.PERSON
+            is AllocationDetail.EventAllocation -> AllocationType.ORDER
+            is AllocationDetail.RequirementAllocation -> AllocationType.REQUIREMENT
+        }
+
+        val allocationReasonCode = when (allocationDetail.allocationReason) {
+            null -> allocationDetail.code
+            else -> deriveDeliusCodeFromTextDefaultInitial(allocationDetail.allocationReason.toString(), allocationType)
+        }
```
marcus-bcl: We're handling null in the `deriveDeliusCodeDefaultInitial` method already - do we need to handle it here as well or can we remove it?

```suggestion
        val allocationReasonCode = deriveDeliusCodeDefaultInitial(allocationDetail.allocationReason, allocationType)
```

PR #5921, opened 2026-01-07
PI-3713 New Limited Access Endpoint for CVL

sambailey-moj:


projects/create-and-vary-a-licence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/resource/UserResource.kt
```diff
@@ -23,4 +28,9 @@ class UserResource(private val ldapTemplate: LdapTemplate) {
     @DeleteMapping(value = ["/{username}/roles"])
     fun removeRole(@PathVariable username: String) =
         ldapTemplate.removeRole(username, DeliusRole.LHDCBT002)
+
+    @PreAuthorize("hasRole('PROBATION_API__CVL__USER_ROLES__RW')")
```
marcus-bcl: Could we change this to use the `PROBATION_API__CVL__CASE_DETAIL` role?  That's the generic one we use for most of the CVL endpoints, the `PROBATION_API__CVL__USER_ROLES__RW` one is a read/write role that allows them to grant/deny access to CVL by modifying Delius user roles.

```suggestion
    @PreAuthorize("hasRole('PROBATION_API__CVL__CASE_DETAIL')")
```

PR #5937, opened 2026-01-12
PI-3739 - Get schedule endpoint for community payback and delius

jamesmacbeth-unilink:


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/entity/UnpaidWorkAppointment.kt
```diff
@@ -314,6 +330,17 @@ interface UnpaidWorkAppointmentRepository : JpaRepository<UpwAppointment, Long>
...
     )
     fun getUpwRequiredAndCompletedMinutes(upwDetailsId: List<Long>): List<UpwMinutesDto>
+
+    @Query(
+        """
+        select a from UpwAppointment a
+        where a.details.disposal.event.id = :eventId
+        and a.softDeleted = false
+        and a.details.softDeleted = false
+        and a.details.disposal.softDeleted = false
```
marcus-bcl: Same here - is it worth moving these soft-deleted/active checks up to the entity?

PR #6003, opened 2026-02-03
PI-3813 new jitbit endpoint

sambailey-moj:
New endpoint to check whether a user has a delius account from their email address

### DIFF
--- a/projects/jitbit-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/User.kt
+++ b/projects/jitbit-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/User.kt
@@ -1,6 +1,0 @@
-package uk.gov.justice.digital.hmpps.model
-
-data class User(
-    val exists: Boolean,
-    val username: String
-)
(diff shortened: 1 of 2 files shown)


projects/jitbit-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/repository/UserRepository.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.justice.digital.hmpps.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import uk.gov.justice.digital.hmpps.entity.User
+import uk.gov.justice.digital.hmpps.exception.NotFoundException
+
+interface UserRepository : JpaRepository<User, Long> {
+    fun findByUserName(userName: String): User?
```
marcus-bcl: Could we use an exists query here instead, to save pulling back the data?
```suggestion
    fun existsByUserName(userName: String): Boolean
```

sambailey-moj:
> We'll need to add the Helm config for the LDAP connection details too.

Is this correct as I've done it ?

projects/jitbit-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/repository/UserRepository.kt
```diff
@@ -0,0 +1,12 @@
...
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
+import uk.gov.justice.digital.hmpps.entity.User
+
+interface UserRepository : JpaRepository<User, Long> {
+    @Query("select u from User u where upper(u.userName) = upper(:username)")
+    fun findUserByUsername(username: String): User?
+}
+
+fun UserRepository.userExists(username: String): String? = findUserByUsername(username)?.userName
```
marcus-bcl: Is it worth renaming this, or changing the body?  It's currently called `userExists` which feels like it should return a Boolean e.g.

```suggestion
fun UserRepository.userExists(username: String) = findUserByUsername(username) != null
```

marcus-bcl (approved):

PR #6005, opened 2026-02-04
PI-3620 - Add Contact endpoint

jamesmacbeth-unilink:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -0,0 +1,78 @@
...
+                    requirement = requirement,
+                    staff = staff,
+                    team = team,
+                    provider = team.provider,
+                    notes = """
+                    ${createContact.notes}
+                    
+                    This contact was automatically created by the Manage Supervision integrations service.
+                """.trimIndent(),
+                    alert = createContact.alert,
```
marcus-bcl: If alert is true, do we also need to create a contact_alert record?

PR #6015, opened 2026-02-06
Add community payback pickup location description

davidatkinsuk:
This commit introduces a common `PickUpLocation` type used by the get appointment schedule endpoints, adding the pickup description.

To achieve this whilst also retaining backwards compatibility we’ve had to introduce a new `CreateAppointmentPickUpData` type and a deprecated `LegacyPickupData` type until usage has been removed by the consumer.


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/PickUpLocation.kt
```diff
@@ -0,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.model
+
+import uk.gov.justice.digital.hmpps.entity.staff.OfficeLocation
+
+data class PickUpLocation(
+    val code: Code,
```
marcus-bcl: Could we make this a string?  That way PickUpLocation is partially JSON-compatible with LegacyPickUp.
```suggestion
    val code: String,
```

It's kind of why we added the Code wrapper class - so that we could add fields like this in a non-breaking way.

marcus-bcl (approved):

PR #6018, opened 2026-02-06
Pi 3750 Basic Details API for COSSO

sambailey-moj:


projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonService.kt
```diff
@@ -0,0 +1,48 @@
...
+                streetName = it.streetName,
+                townCity = it.townCity,
+                district = it.district,
+                county = it.county,
+                postcode = it.postcode,
+            )
+        }
+        if (addresses.none { it.status.uppercase() == MAIN_ADDRESS_STATUS }) {
+            throw NotFoundException("No main address found for person with crn:$crn")
+        }
```
marcus-bcl: Why do we need to check if they have a main address?  I can't see anything about that in the story.

projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/PersonService.kt
```diff
@@ -0,0 +1,50 @@
...
+                postcode = it.postcode,
+            )
+        }
+        return PersonDetails(
+            title = person.title!!.description,
+            name = Name(person.firstName, person.middleName(), person.surname),
+            dateOfBirth = person.dateOfBirth,
+            telephoneNumber = person.telephoneNumber ?: "",
+            mobileNumber = person.mobileNumber ?: "",
+            emailAddress = person.emailAddress ?: "",
```
marcus-bcl: Would it be better to just return nulls here?
```suggestion
            title = person.title?.description,
            name = Name(person.firstName, person.middleName(), person.surname),
            dateOfBirth = person.dateOfBirth,
            telephoneNumber = person.telephoneNumber,
            mobileNumber = person.mobileNumber,
            emailAddress = person.emailAddress,
```

marcus-bcl (approved):

PR #6048, opened 2026-02-11
PI-3752 COSSO Offence Details Endpoint

sambailey-moj:


projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/AdditionalOffence.kt
```diff
@@ -0,0 +1,27 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.Table
+import org.springframework.data.jpa.repository.JpaRepository
+import java.time.LocalDate
+
+@Entity
+@Table(name = "additional_offence")
+class AdditionalOffence(
```
marcus-bcl: Should we filter on soft_deleted?

Also, minor thing - you don't need the `@Table` annotation when it matches the entity name

projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/MainOffence.kt
```diff
@@ -0,0 +1,46 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.ManyToOne
+import jakarta.persistence.OneToOne
+import jakarta.persistence.Table
+import org.springframework.data.jpa.repository.JpaRepository
+
+@Entity
+@Table(name = "main_offence")
```
marcus-bcl: Same here - filter on soft_deleted

projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/OffenceService.kt
```diff
@@ -0,0 +1,83 @@
...
+    fun getOffenceDetails(uuid: String): OffenceDetails {
+
+        val eventId = documentRepository.findEventIdFromDocument(
+            DocumentEntity.cossoBreachNoticeUrn(UUID.fromString(uuid))
+        )
+            ?: throw NotFoundException("DocumentEntity", "UUID", uuid)
+        val mainOffence = mainOffenceRepository.findByEventId(eventId)?.offence
+            ?: throw NotFoundException("Offence", "eventId", eventId)
+        val additionalOffences = additionalOffenceRepository.findAllByEventId(eventId)
+            .map { CodedDescription(it.offence.mainCategoryCode, it.offence.mainCategoryDescription) }
```
marcus-bcl: Sorry, I didn't spot this earlier - I believe it's the subCategory code and description we want to return for both main and additional offences.  The main category is the high-level grouping, and the sub category is the actual offence code.  I don't think we need to map the main category fields on the offence entity at all.

PR #6066, opened 2026-02-18
PI-3750 Add Responsible Officer endpoint for COSSO

sambailey-moj:


projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ResponsibleOfficerService.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    private fun getResponsibleOfficerStaff(
+        offenderManager: OffenderManager?,
+        prisonOffenderManager: PrisonOffenderManager?,
+        responsibleOfficer: ResponsibleOfficer
+    ): Staff {
+        val staff = offenderManager?.staff ?: prisonOffenderManager?.staff
+        if (staff == null) throw NotFoundException("Staff", "responsibleOfficerId", responsibleOfficer.id)
+        return staff
+    }
```
marcus-bcl: Minor thing - do you need to pass all three entities here, or just the `ResponsibleOfficer`?  It might simplify things to make this one and `getProbationAreaForResponsibleOfficer` into extension functions of `ResponsibleOfficer` e.g.
```suggestion
    private fun ResponsibleOfficer.getStaff() = 
        (offenderManager?.staff ?: prisonOffenderManager?.staff).orNotFoundBy("crn", person.crn)
```


Then you can just call them above as:
```kotlin
val staff = responsibleOfficer.getStaff()
val probationArea = responsibleOfficer.getProbationArea()
```

marcus-bcl (commented):
Looks good!

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #6070, opened 2026-02-19
PI-3861 unpaid work details summary

sambailey-moj:


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseSummaryService.kt
```diff
@@ -0,0 +1,50 @@
...
+class CaseSummaryService(
+    private val personRepository: PersonRepository,
+    private val eventRepository: EventRepository,
+    private val upwDetailsRepository: UpwDetailsRepository,
+    private val unpaidWorkAppointmentRepository: UnpaidWorkAppointmentRepository,
+    private val linkedListRepository: LinkedListRepository
+) {
+    fun getSummaryForCase(crn: String): UnpaidWorkDetails {
+        val personId = personRepository.getByCrn(crn).id
+        val events = eventRepository.getByPerson_Id(personId)
```
marcus-bcl: Is it worth combining these into one lookup?
```suggestion
        val events = eventRepository.getByPersonCrn(crn)
```

marcus-bcl (approved):

PR #6083, opened 2026-02-24
retries for alfrescoClient

sambailey-moj:


libs/document-management/build.gradle.kts
```diff
@@ -5,6 +5,7 @@ dependencies {
     implementation(project(":libs:audit"))
     implementation("org.springframework.boot:spring-boot-starter-webmvc")
     implementation("org.springframework.boot:spring-boot-starter-security")
+    implementation("org.springframework.retry:spring-retry:1.3.4")
```
marcus-bcl: It looks like Spring Retry is no longer actively being developed: https://github.com/spring-projects/spring-retry

libs/document-management/src/main/kotlin/uk/gov/justice/digital/hmpps/client/AlfrescoClient.kt
```diff
@@ -16,6 +19,21 @@ import java.util.*
 class AlfrescoClient(
     @Qualifier("alfrescoRestClient") private val restClient: RestClient
 ) {
+    // RetryTemplate for GET calls
+    private val retryTemplate: RetryTemplate = RetryTemplate().apply {
```
marcus-bcl: We already have our own retry function in `Retry.kt`. I'm not against replacing it with a library if you want, but should we do that everywhere - so we don't have multiple parallel retry implementations?

PR #6114, opened 2026-03-03
FM-354 AP & Delius - Add support for personal contacts in probation case entities.

danhumphreys-moj:
**Summary**

This Pull Request introduces changes to enhance the handling of personal contact information for probation cases in the application. It adds entities and structures to support data persistence and retrieval of personal contact details.

The implementation is based on similar personal contact logic in the 'unpaid-work-and-delius' project: 
https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/ea14f7928cd3963ae17a676feb80b2ee2a0fbc1d/projects/unpaid-work-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/casedetails/entity/CasePersonalContactEntity.kt#L18

**Main Changes:**

Modification of `ProbationCase`:

- Added a new field, `personalContacts`, to store related personal contact entities.

Addition of New Entities:

- `AddressEntity`: Represents address information.
- `PersonalContactEntity`: Represents personal contact details linked to a probation case.

Enhancement of `CaseDetail`:

- Introduced a new `personalContacts` field to include detailed contact information in the case details.
- Defined supporting classes for structured handling of contact, address, and relationship details.

Changes to `CaseService`:

- Updated methods to handle retrieval and mapping of personal contact data into case summaries and details.
- These changes aim to extend functionality by including personal contact management while preserving existing structures

### DIFF
--- a/projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseDetail.kt
+++ b/projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseDetail.kt
@@ -25,7 +25,32 @@ data class CaseDetail(
     val mappaDetail: MappaDetail?,
     val careLeaver: Boolean,
     val veteran: Boolean,
-    val sentences: List<Sentence>
+    val sentences: List<Sentence>,
+    val personalContacts: List<PersonalContact>? = listOf(),
+)
+
+data class PersonalContact(
+    val relationship: String,
+    val relationshipType: RelationshipType,
+    val name: Name,
+    val telephoneNumber: String?,
+    val mobileNumber: String?,
+    val address: Address?
+)
+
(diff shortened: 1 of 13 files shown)


projects/approved-premises-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseDetail.kt
```diff
@@ -25,7 +25,38 @@ data class CaseDetail(
...
+data class RelationshipType(
+    val code: String,
+    val description: String
+)
+
+data class ContactName(
+    val forename: String,
+    val middleName: String?,
+    val surname: String
+)
```
marcus-bcl: Would it be worth reusing the existing `Name` or `PersonName` model here?  We should probably only have one definition of name, not sure how we've ended up with two already 😄

PR #6268, opened 2026-04-10
PI-4023 Add sentence type to endpoints

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AppointmentService.kt
```diff
@@ -43,6 +47,18 @@ class AppointmentService(
...
+            .mapValues { (_, appearances) ->
+                when (appearances.maxWithOrNull(
+                    compareBy<CourtAppearance>(
+                        { it.date },
+                        { it.type.code })
+                )?.type?.code) {
+                    "S" -> SentenceType.COMMUNITY
+                    else -> SentenceType.PRE_SENTENCE
+                }
+            }
```
marcus-bcl: Could we check for the court appearance type in the SQL query, to avoid pulling back every appearance record?

E.g. something like this would give you a map of Event ID to a boolean indicating whether they have a sentencing court appearance:
```suggestion
        val eventsWithSentencingCourtAppearance = courtAppearanceRepository.findIdByEventInAndTypeCode(activeEvents, "S")
            .groupBy { it.event.id }
            .mapValues { (_, ids) -> ids.isNotEmpty() }

```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceService.kt
```diff
@@ -89,18 +91,33 @@ class SentenceService(
...
-            id,
-            eventNumber,
-            disposal?.toMinimalOrder(),
-            licenceConditions = disposal?.let {
-                licenceConditionRepository.findAllByDisposalId(disposal.id).asMinimals()
-            } ?: emptyList(),
-            requirements = requirementRepository.getRequirements(id, eventNumber, includeRarRequirements).asMinimals {
-                requirementService.getRar(it.disposal!!.id, it.mainCategory!!.code)
+    fun List<Event>.toMinimalSentences(includeRarRequirements: Boolean): List<MinimalSentence> {
+        val sentencingCourtAppearance = courtAppearanceRepository.getCourtAppearancesByEventInAndType_Code(this, "S")
```
marcus-bcl: Is this a duplicate of the method in AppointmentService?  Is there a way we can make this into a common extension function that can be shared between the two services?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/sentence/entity/SentenceRepository.kt
```diff
@@ -27,6 +28,8 @@ interface EventSentenceRepository : JpaRepository<Event, Long> {
 }
 
 interface CourtAppearanceRepository : JpaRepository<CourtAppearance, Long> {
+    @EntityGraph(attributePaths = ["event", "type"])
+    fun getCourtAppearancesByEventInAndType_Code(events: List<Event>, typeCode: String): List<CourtAppearance>
```
marcus-bcl: I guess this returns the full entity for each court appearance?  If we only need the Event IDs, is it possible to only return that field?

```suggestion
    @Query("select c.event.id from CourtAppearance c where c.event in (:events) and c.type.code = 'S'")
    fun findEventIdsWithSentencingCourtAppearances(events: List<Event>): List<Long>
```

marcus-bcl (dismissed):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/SentenceService.kt
```diff
@@ -55,9 +55,11 @@ class SentenceService(
         val person = personRepository.getPerson(crn)
         val activeEvents = getActiveSentences(person.id)
 
+        activeEvents.toMinimalSentences(includeRarRequirements)
```
marcus-bcl: Is this intentional?

marcus-bcl (approved):

PR #6291, opened 2026-04-17
PI-4038 - Update sign-and-send endpoint to return empty address list instead of 404

jamesmacbeth-unilink:


sambailey-moj (dismissed):

projects/suicide-risk-form-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/DetailsService.kt
```diff
@@ -40,22 +40,23 @@ class DetailsService(
...
+            val ldapUser = ldapTemplate.findByUsername<LdapUser>(user.username)
+            val defaultReplyAddress =
+                ldapTemplate.findPreferenceByUsername(user.username, "replyAddress")?.toLongOrNull()
+            val addresses = if (ldapUser?.userHomeArea != null) {
+                officeLocationRepository.findAllByProviderCode(ldapUser.userHomeArea).map {
+                    it.toAddress().copy(status = if (it.id == defaultReplyAddress) "Default" else null)
+                }
+            } else emptyList()
+            Pair(ldapUser, addresses)
+        } ?: Pair(null, emptyList())
```
marcus-bcl: Do we need to use a Pair here?  Might be simpler to get the LDAP user first, then get the addresses if the LDAP user has a home area?

e.g. something like:

```suggestion
        val ldapUser = userRepository.findByStaffId(staff.id)
            ?.let { ldapTemplate.findByUsername<LdapUser>(it.username) }
        val addresses = ldapUser?.userHomeArea?.let { homeArea ->
	        val defaultReplyAddress = 
	            ldapTemplate.findPreferenceByUsername(ldapUser.username, "replyAddress")?.toLongOrNull()
            officeLocationRepository.findAllByProviderCode(homeArea).map {
                it.toAddress().copy(status = if (it.id == defaultReplyAddress) "Default" else null)
            }
        } ?: emptyList()
```

marcus-bcl (approved):

PR #6292, opened 2026-04-17
PI-4030 Handle appointments linked to inactive requirements

marcus-bcl:
Also, updated entity graphs to optimise generated SQL queries


sambailey-moj (approved):
Sonar failing..

PR #6293, opened 2026-04-17
PI-4035 - Add PDUs to MPOP user roles endpoint

jamesmacbeth-unilink:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -77,7 +79,10 @@ class UserService(
         surname = surname,
         email = email,
         enabled = enabled,
-        roles = getUserRoles(dn)
+        roles = getUserRoles(dn),
+        staff = getUserPdus(userId)?.let { pdus ->
+            StaffDetails(pdus.map { ProbationDeliveryUnit(it.code, it.description) })
+        }
     )
```
marcus-bcl: Are we fetching the user twice here?  Once to get the ID by username, then using the ID to fetch again and get the staff.  Could we just pass the full user entity instead of just the ID to LdapUser.toUserDetails, then get the staff from there?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -88,6 +93,10 @@ class UserService(
         AttributesMapper { it["cn"].get().toString() }
     )
 
+    private fun getUserPdus(userId: Long) = userRepository.findUserById(userId).staff?.let {
+        boroughRepository.findAllByProvider_Id(it.provider.id)
+    }
```
marcus-bcl: I think we need the list of PDUs (boroughs) linked to each of the teams, rather than linked to the staff provider.

```
staff -> team* -> district -> borough -> probation_area
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Contact.kt
```diff
@@ -990,5 +990,11 @@ class Borough(
 )
 
 interface BoroughRepository : CrudRepository<Borough, Long> {
```
marcus-bcl: Should this be a JpaRepository ?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Contact.kt
```diff
@@ -980,7 +980,21 @@ class Borough(
...
\ No newline at end of file
+)
+
+interface BoroughRepository : CrudRepository<Borough, Long> {
+    @Query(
+        "select distinct t.district.borough " +
+            "from Team t " +
+            "join t.staff s " +
+            "where s.id = :staffId"
+    )
```
marcus-bcl, line 1000: Minor thing - could we use a multi-line string here?
```suggestion
    @Query(
        """
        select distinct t.district.borough
        from Team t
        join t.staff s
        where s.id = :staffId
        """
    )
```

marcus-bcl (approved):

PR #6294, opened 2026-04-20
PI-4011 Add single case endpoint

hazzaz-justice:
Branched out from PI-4011 and verified the commits


marcus-bcl (changes requested):
@hazzaz-justice please could you rebase or merge main into this branch? The PR is currently showing a number of other changes currently on main.

projects/single-accommodation-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseService.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.exception.NotFoundException.Companion.orNotFoundBy
+import uk.gov.justice.digital.hmpps.integrations.delius.*
+import uk.gov.justice.digital.hmpps.model.*
+
+@Service
+@Transactional(readOnly = true)
+class CaseService(
```
marcus-bcl: There appears to be some duplication between this class and the existing `CaseListService` - is there a reason we can't merge the two into a single `CaseService` with two methods (e.g. `getCase` and `getCaseList`)?

hazzaz-justice: Done, please have a look @marcus-bcl

projects/single-accommodation-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/PersonManager.kt
```diff
@@ -37,4 +37,5 @@ class PersonManager(
 
 interface PersonManagerRepository : JpaRepository<PersonManager, Long> {
     fun findByStaffIdAndActiveTrue(staffId: Long): List<PersonManager>
+    fun findFirstByPersonIdAndActiveTrueAndSoftDeletedFalse(personId: Long): PersonManager?
```
marcus-bcl: Could we move the soft-deleted check into a `SqlRestriction` on the entity?  It looks like it was missed on the other endpoint.

We should be able to add `@SqlRestriction("soft_deleted = 0 and active_flag = 1")` to the `PersonManager` entity, then remove the `AndActiveTrue...` from these two methods.

projects/single-accommodation-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseListService.kt
```diff
@@ -70,4 +71,50 @@ class CaseListService(
...
+                description = manager.team.description
+            ),
+            gender = person.gender.description,
+            roshLevel = roshLevel,
+            expectedReleaseDate = keyDateRepository.findExpectedReleaseDates(person.id),
+            userExcluded = access.userExcluded,
+            userRestricted = access.userRestricted,
+            exclusionMessage = access.exclusionMessage,
+            restrictionMessage = access.restrictionMessage
+        )
```
marcus-bcl: This bit looks nearly identical to the one in the getCaseList method - could we extract it to a shared private function in the CaseListService?  It'd be good to ensure we're returning a consistent view of the case, whether it's individually or in a list.

marcus-bcl (approved):

PR #6313, opened 2026-04-22
PI-4037 - Initial endpoints for Probation Access Control service

jamesmacbeth-unilink:


projects/probation-access-control/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/generator/UserGenerator.kt
```diff
@@ -1,7 +1,11 @@
...
 
+import uk.gov.justice.digital.hmpps.entity.LimitedAccessUser
 import uk.gov.justice.digital.hmpps.user.AuditUser
 
 object UserGenerator {
     val AUDIT_USER = AuditUser(IdGenerator.getAndIncrement(), "ProbationAccessControl")
+
+    val DEFAULT = LimitedAccessUser("officer", IdGenerator.getAndIncrement())
+    val RESTRICTED = LimitedAccessUser("restricted", IdGenerator.getAndIncrement())
 }
```
sambailey-moj, line 11: Is it worth having an 'Excluded' user too. Would we want to test a user with no exclusions or restrictions, a user with a exclusion(s), and a user with restriction(s)

marcus-bcl (approved):

PR #6322, opened 2026-04-28
PI-4075 fix mpop bug with misaligned home area and team

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -266,7 +266,7 @@ class UserService(
         return DefaultUserDetails(
             default.username,
             default.staff?.code,
-            providers.first { it.code == homeArea }.name,
+            providers.firstOrNull { it.code == homeArea }?.name.orNotFoundBy("code", homeArea),
```
marcus-bcl: Would it be better to just look up their home provider from the DB if it's not available in the providers list?

sambailey-moj: The problem we've encountered is when their home provider differs from their team's provider. This happens more than you might imagine! For the purpose of patching this bug - throwing a 404 instead of a 500 - would this fix be sufficient. Perhaps we could go back to MPoP & ask them what they would like as a longer term fix - and change the behaviour in another ticket?

marcus-bcl (approved):

PR #6323, opened 2026-04-29
PI-4069 - Add paging and optimise SAS case list endpoint queries

jamesmacbeth-unilink:


projects/single-accommodation-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/CaseListResponse.kt
```diff
@@ -3,7 +3,11 @@ package uk.gov.justice.digital.hmpps.model
 import java.time.LocalDate
 
 data class CaseListResponse(
-    val cases: List<Case>
+    val cases: List<Case>,
+    val totalElements: Long,
+    val totalPages: Int,
+    val page: Int,
+    val size: Int,
```
marcus-bcl: Should we move these metadata fields to their own object, to keep it consistent with Spring's `PagedModel` type?

E.g.

```suggestion
    val cases: List<Case>,
    val page: org.springframework.data.web.PagedModel.PageMetadata,
```

projects/single-accommodation-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseListService.kt
```diff
@@ -1,80 +1,76 @@
...
+        val cases = casesPageable.content
+        val personIds = cases.map { it.id }
+        val limitedAccess =
+            userAccessService.userAccessFor(username, cases.map { it.crn }).access.associateBy { it.crn }
+        val expectedReleaseDates = keyDateRepository.findExpectedReleaseDatesByPersonIdIn(personIds)
+            .associate { it.personId to it.releaseDate }
         val roshLevels = registrationRepository.findByPersonIdInAndTypeCodeIn(personIds, RegisterType.ROSH_CODES)
-            .groupBy { it.personId }
+            .groupBy { it.person?.id }
             .mapValues { (_, reg) -> CodeDescription(reg.first().type.code, reg.first().type.description) }
```
marcus-bcl: Do we still need this if we're mapping `roshRegistrations` on the `Person` entity?

projects/single-accommodation-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseListService.kt
```diff
@@ -1,80 +1,84 @@
...
+        val expectedReleaseDates = keyDateRepository.findExpectedReleaseDatesByPersonIdIn(personIds)
+            .associate { it.personId to it.releaseDate }
+        val roshLevels = cases.associate { reg ->
+            reg.id to reg.roshRegistrations.firstOrNull()?.type?.let {
+                CodeDescription(
+                    it.code,
+                    it.description
+                )
+            }
+        }
```
marcus-bcl: It doesn't matter too much, but given we're eagerly fetching these via the entity graph - could we just do this mapping in the `Person.toCase` method?
```kotlin
       roshLevel = roshRegistrations.firstOrNull()?.type?.let { CodeDescription(it.code, it.description) }
```

projects/single-accommodation-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseListService.kt
```diff
@@ -1,80 +1,75 @@
...
-        val roshLevel = registrationRepository
-            .findByPersonIdInAndTypeCodeIn(listOf(person.id), RegisterType.ROSH_CODES)
-            .firstOrNull()
-            ?.let { CodeDescription(it.type.code, it.type.description) }
-
         val access = userAccessService.caseAccessFor(username, crn)
-
-        return toCase(person, manager, access, roshLevel)
+        val roshLevel = person.roshRegistrations.firstOrNull()
+            ?.let { CodeDescription(it.type.code, it.type.description) }
```
marcus-bcl: Can we remove this line?

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #6344, opened 2026-05-05
PI-4087 - Update Create Contact endpoint to add outcomes and enforcement actions

jamesmacbeth-unilink:


projects/manage-supervision-and-delius/deploy/database/access.yml
```diff
@@ -10,6 +10,7 @@ database:
       - audited_interaction
       - document
       - enforcement
+      - r_enf_act_contact_out_type
```
sambailey-moj, line 13: Do you need this? Are we writing to the `r_enf_act_contact_out_type` table?

jamesmacbeth-unilink: Good catch, we won't be writing to it so don't need this. Thanks!

sambailey-moj (approved):

PR #6351, opened 2026-05-07
PI-4090 - Add contact suspended flag to esupervision api

jamesmacbeth-unilink:


projects/esupervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Registration.kt
```diff
@@ -0,0 +1,56 @@
...
+    companion object {
+        const val CONTACT_SUSPENDED_TYPE_CODE = "PRC"
+    }
+}
+
+interface RegistrationRepository : JpaRepository<Registration, Long> {
+    fun existsByPersonIdAndTypeCode(personId: Long, typeCode: String): Boolean
+
+    @Query("select r.personId from Registration r where r.personId in :personIds and r.type.code = :typeCode")
+    fun findPersonIdsWithActiveType(personIds: Collection<Long>, typeCode: String): List<Long>
```
marcus-bcl: Could we just return a `Set` here, to save the `toSet` call in ContactDetailsService?

marcus-bcl (approved):

PR #6354, opened 2026-05-11
PI-4098 DPS And Delius - restrict any record with LAO details

sambailey-moj:


projects/dps-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Document.kt
```diff
@@ -173,6 +173,6 @@ interface DocumentRepository : JpaRepository<DocumentEntity, Long> {
     )
     fun getPersonAndEventDocuments(personId: Long): List<Document>
 
-    @Query("select d.name from DocumentEntity d where d.alfrescoId = :alfrescoId")
-    fun findNameByAlfrescoId(alfrescoId: String): String?
+    @Query("select d from DocumentEntity d where d.alfrescoId = :alfrescoId")
```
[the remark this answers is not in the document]
sambailey-moj: This is the existing logic - we are not looking to change this in this PR

projects/dps-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/DocumentService.kt
```diff
@@ -17,58 +18,72 @@ import uk.gov.justice.digital.hmpps.model.ProbationDocumentsResponse
...
+        } ?: throw NotFoundException("Person", "nomisId", nomisId)
 
     private val Disposal.description get() = "${type.description}${lengthString?.let { " ($it)" } ?: ""}"
     private val Disposal.lengthString get() = length?.let { "$length ${lengthUnits!!.description}" }
     private fun List<CourtAppearance>.latestOutcome() = filter { it.outcome != null }.maxByOrNull { it.date }?.outcome
+    private fun checkForLao(crn: String, username: String) {
+        val limitedAccessInfo = limitedAccessService.checkLimitedAccessFor(listOf(crn))
+        limitedAccessInfo.access.firstOrNull { it.crn == crn }?.let { limitedAccess ->
+            if (limitedAccess.userExcluded || limitedAccess.userRestricted) {
+                throw AccessDeniedException("Access Denied for user $username to case with crn ${crn}")
```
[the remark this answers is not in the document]
sambailey-moj: This is intentional logic

projects/dps-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/DocumentService.kt
```diff
@@ -17,58 +18,72 @@ import uk.gov.justice.digital.hmpps.model.ProbationDocumentsResponse
...
+                        }?.sortedByDescending { it.createdAt } ?: emptyList()
+                    )
+                }.sortedByDescending { it.date }
+            )
+        } ?: throw NotFoundException("Person", "nomisId", nomisId)
 
     private val Disposal.description get() = "${type.description}${lengthString?.let { " ($it)" } ?: ""}"
     private val Disposal.lengthString get() = length?.let { "$length ${lengthUnits!!.description}" }
     private fun List<CourtAppearance>.latestOutcome() = filter { it.outcome != null }.maxByOrNull { it.date }?.outcome
+    private fun checkForLao(crn: String, username: String) {
```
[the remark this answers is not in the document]
sambailey-moj: username is just for logging purposes, this is intentional

projects/dps-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/DocumentController.kt
```diff
@@ -19,9 +20,11 @@ class DocumentController(private val documentService: DocumentService) {
             to in the probation case (e.g. a court appearance, an event, etc).
         """
     )
-    fun getDocuments(@PathVariable nomisId: String) = documentService.getDocumentsForCase(nomisId)
+    fun getDocuments(@PathVariable nomisId: String, authentication: Authentication) =
+        documentService.getDocumentsForCase(nomisId, authentication.name)
```
marcus-bcl: I don't think we want to pass the username through here?  DPS is a prison service, so the username provided will be a NOMIS/DPS username rather than a Delius username.

We want to check if the case is restricted to _any_ user, or excluded from _any_ user - as we can't tell whether the DPS user is the same person as one of the Delius users who does not have access.

sambailey-moj: Ah but I think it will be useful to pass the username to log.. So in case of any audit/investigation there is a log of the user (albeit a NOMIS ID) who tried to access the record

marcus-bcl: Actually I don't think you'll even get the NOMIS username, I expect it'll just be the Client ID (e.g. "digital-prison-services") given it's a client_credentials access token.

projects/dps-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/DocumentService.kt
```diff
@@ -17,58 +18,72 @@ import uk.gov.justice.digital.hmpps.model.ProbationDocumentsResponse
...
+    private val limitedAccessService: UserAccessService,
 ) {
     fun downloadDocument(id: String): ResponseEntity<StreamingResponseBody> {
-        val filename =
-            documentRepository.findNameByAlfrescoId(id) ?: throw NotFoundException("Document", "alfrescoId", id)
-        return alfrescoClient.streamDocument(id, filename)
+        val document =
+            documentRepository.findByAlfrescoId(id) ?: throw NotFoundException("Document", "alfrescoId", id)
+        val person = personRepository.findById(document.personId)
+            .orElseThrow { NotFoundException("Person", "id", document.personId) }
```
marcus-bcl, line 28: It's a bit heavyweight to pull back the full document and person record to only use the CRN, could it be improved by adding a `documentRepository.findCrnByAlfrescoId(id)` method (or something like that)?

Not the end of the world though, happy to approve as-is.

marcus-bcl (approved):

PR #6356, opened 2026-05-11
PI-4044 adding in enforcement action to PersonAppointment for mpop1

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ScheduleService.kt
```diff
@@ -30,13 +31,15 @@ class ScheduleService(
...
-    private val userRepository: UserRepository
+    private val userRepository: UserRepository,
+    private val enforcementRepository: EnforcementRepository,
 ) {
 
     fun getPersonAppointment(crn: String, contactId: Long, noteId: Int? = null): PersonAppointment {
         val summary = personRepository.getSummary(crn)
         val contact = contactRepository.getContact(summary.id, contactId)
         val documents = contact.documents
+        val enforcement = enforcementRepository.findFirstByContactIdOrderByIdDesc(contactId)
```
marcus-bcl: A contact can only have 0 or 1 enforcement entities linked to it.  Is there any reason not to model this as a simple `@OneToOne` join column on the Contact entity, to avoid needing an additional query?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/compliance/EnforcementAction.kt
```diff
@@ -0,0 +1,9 @@
+package uk.gov.justice.digital.hmpps.api.model.compliance
+
+import java.time.LocalDate
+
+data class EnforcementAction(
+    val code: String?,
+    val description: String?,
```
marcus-bcl, line 7: Do these two need to be nullable, if the entire `PersonAppointment.enforcementAction` field can be null?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ScheduleService.kt
sambailey-moj: done - can you please check the mappings on both entities look ok?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/api/model/compliance/EnforcementAction.kt
sambailey-moj: fixed

marcus-bcl (approved):

PR #6363, opened 2026-05-12
PI-4093 assessment summary change concerns fields

sambailey-moj:


projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/oasys/OrdsClient.kt
```diff
@@ -66,17 +64,15 @@ data class AssessmentSummary(
 
     /*
         Order of Concerns has to match Delius read order:
-        Suicide, Harm, Custody, Hostel, Vulnerability, Abscond, Disruptive Behaviour, Breach of Trust
+        Suicide, SelfHarm, Custody, Vulnerability, Abscond, Behave Trust
```
marcus-bcl: These values haven't changed in Delius, so we must maintain the mapping here:
```suggestion
        Suicide, Harm, Custody, Hostel, Vulnerability, Abscond, Disruptive Behaviour, Breach of Trust
```

```
val concernFlags: List<String> = listOf(
	concernsRiskOfSuicide,
	concernsRiskOfSelfHarm,
	concernsCopingInCustody, // intentionally duplicated
	concernsCopingInCustody, // intentionally duplicated
	concernsVulnerability,
	concernsEscapeAscond,
	currentControlBehaveTrust, // intentionally duplicated
	currentControlBehaveTrust, // intentionally duplicated
```

> The “concernFlags” are stored in Delius as a comma-separated string of YES/NO/DK strings, so the order must be preserved.

projects/assessment-summary-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/assessment/entity/OasysAssessment.kt
```diff
@@ -144,7 +144,7 @@ class OasysAssessment(
 
     companion object {
         val RISK_FLAGS_PATTERN: Regex = "^([VvHhMmLlNn],){8}[VvHhMmLlNn]$".toRegex()
-        val CONCERN_FLAGS_PATTERN: Regex = "^((NO|YES|DK),){7}(NO|YES|DK)$".toRegex()
+        val CONCERN_FLAGS_PATTERN: Regex = "^(((NO|YES|DK),){5}(NO|YES|DK)|((NO|YES|DK),){7}(NO|YES|DK))$".toRegex()
```
marcus-bcl: This pattern shouldn't change - the resulting value we're storing in Delius needs to stay the same.

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #6390, opened 2026-05-18
Pi 4131 enforcement action contact creation mpop

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +137,37 @@ class ContactLogService(
...
+                // increment ftcCount using logic from appts library
+                contactRepository.countFailureToComply(event!!).takeIf { it > 0 }?.let { ftcCount ->
+                    event.ftcCount = ftcCount
+                    eventRepository.save(event)
+                }
+                savedContact.event?.run {
+                    ftcCount = (ftcCount ?: 0) + 1
+                    val ftcLimit = event.disposal?.type?.ftcLimit
+                    contactRepository.save(savedContact)
+                    if ((ftcCount ?: 1) > (ftcLimit ?: 0)) {
```
marcus-bcl: The FTC_COUNT column is non-nullable, so we shouldn't need this fallback.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +137,37 @@ class ContactLogService(
...
                     )
                 )
-                savedContact.event?.run { ftcCount = (ftcCount ?: 0) + 1 }
+                // increment ftcCount using logic from appts library
+                contactRepository.countFailureToComply(event!!).takeIf { it > 0 }?.let { ftcCount ->
+                    event.ftcCount = ftcCount
+                    eventRepository.save(event)
+                }
+                savedContact.event?.run {
+                    ftcCount = (ftcCount ?: 0) + 1
```
marcus-bcl: We don't need to increment again here, we've already calculated the correct value in the previous step

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +137,37 @@ class ContactLogService(
                         """.trimIndent()
                     )
                 )
-                savedContact.event?.run { ftcCount = (ftcCount ?: 0) + 1 }
+                // increment ftcCount using logic from appts library
+                contactRepository.countFailureToComply(event!!).takeIf { it > 0 }?.let { ftcCount ->
```
marcus-bcl: The result will always be positive, so no need for this `.takeIf { it > 0 }?.let`.

Also, you might as well move this into the `savedContact.event?.run { ...` block - otherwise the `event!!` will give you a null-pointer exception for non-event contacts.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
marcus-bcl: Also, by defaulting the ftcLimit to 0, you're effectively saying that an enforcement review should always be created - but it's the opposite that we want.  See: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/0ea97df4d932d11915ad878df5df25b38c4197a2/libs/appointments/src/main/kotlin/uk/gov/justice/digital/hmpps/appointments/service/EnforcementService.kt#L59-L62
(it returns instead of continuing when there is no limit)

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +137,37 @@ class ContactLogService(
...
+                                startTime = ZonedDateTime.now(EuropeLondon),
+                                person = person,
+                                event = event,
+                                staff = staff,
+                                team = team,
+                                provider = team.provider,
+                                notes = """
+                                    ${DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm").format(LocalDateTime.now())}
+                                    Auto-generated enforcement review due to failure to comply count exceeding limit of $ftcLimit.
+                                """.trimIndent()
```
marcus-bcl: Where did this text come from?  I don't think we should be populating enforcement review contact notes.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +137,37 @@ class ContactLogService(
...
+                    eventRepository.save(event)
+                }
+                savedContact.event?.run {
+                    ftcCount = (ftcCount ?: 0) + 1
+                    val ftcLimit = event.disposal?.type?.ftcLimit
+                    contactRepository.save(savedContact)
+                    if ((ftcCount ?: 1) > (ftcLimit ?: 0)) {
+                        val reviewType = contactTypeRepository.findByCode(enforcementAction.contactType.code)
+                            ?: throw NotFoundException("ContactType", "code", enforcementAction.contactType.code)
+                        contactRepository.save(
```
marcus-bcl: We need to check whether an enforcement review already exists for the event before creating a new one - see https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/0ea97df4d932d11915ad878df5df25b38c4197a2/libs/appointments/src/main/kotlin/uk/gov/justice/digital/hmpps/appointments/repository/AppointmentRepositories.kt#L29-L42

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +140,31 @@ class ContactLogService(
                         """.trimIndent()
                     )
                 )
-                savedContact.event?.run { ftcCount = (ftcCount ?: 0) + 1 }
+                savedContact.event?.run {
+                    event?.ftcCount = ++ftcCount
+                    val ftcLimit = event?.disposal?.type?.ftcLimit ?: return@run
```
marcus-bcl: This should use the `countFailureToComply` repository method, rather than incrementing:
```suggestion
                    ftcCount = contactRepository.countFailureToComply(this)
                    val ftcLimit = disposal?.type?.ftcLimit ?: return@run
```


Also, you're referencing the nullable `event`, but `this` refers the non-nullable event in the [run](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/run.html) scope.  So you can just use `this` instead of `event`, and remove the extra null-checks.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +140,31 @@ class ContactLogService(
...
+                    contactRepository.save(savedContact)
```
marcus-bcl: You don't need to save this again.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +140,31 @@ class ContactLogService(
...
+                    val enforcementReviewDoesNotExist = event?.let {
+                        !contactRepository.enforcementReviewExists(it.id, it.breachEnd, REVIEW_ENFORCEMENT_STATUS)
+                    } ?: true
```
marcus-bcl: Same here, if you use the `run` parameter (`this`) then you don't need the null-check:

```suggestion
                    val enforcementReviewDoesNotExist = !contactRepository.enforcementReviewExists(id, breachEnd, REVIEW_ENFORCEMENT_STATUS)
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -137,7 +140,31 @@ class ContactLogService(
...
+                    if (ftcCount > ftcLimit && enforcementReviewDoesNotExist) {
```
marcus-bcl: The line above this will always make a call to the database even when the FTC count is less than the limit.  It might be worth moving it into the if-statement to benefit from short-circuit evaluation.
```suggestion
                    if (ftcCount > ftcLimit && !contactRepository.enforcementReviewExists(...)) {
```


If you really wanted to keep it as a separate variable, you could use `lazy` e.g.
```kotlin
val enforcementReviewDoesNotExist = lazy { !contactRepository.enforcementReviewExists(...) }
if (ftcCount > ftcLimit && enforcementReviewDoesNotExist.value) {

File: projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt [hunk 10]
```diff
@@ -0,0 +1,105 @@
...
+        contactRepository.save(
+            Contact(
+                person = person,
+                event = event,
+                requirement = requirement,
+                type = enforcementAction.contactType,
+                date = LocalDate.now(),
+                startTime = ZonedDateTime.now(EuropeLondon),
+                staff = staff,
+                team = team,
```
marcus-bcl: I don't think we need the bit at the top to fetch these, as we already have them from the contact?
```suggestion
                staff = contact.staff,
                team = contact.team,
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,105 @@
...
+                provider = team.provider,
+                linkedContactId = contact.id,
+                notes = """
+                            ${DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm").format(LocalDateTime.now())}
+                            Enforcement Action: ${enforcementAction.description}
+                        """.trimIndent()
+            )
+        )
+        event?.run {
+            ftcCount = ftcCount?.plus(1)
```
marcus-bcl: We've gone back to incrementing this manually, this should be recalculated using the repository method.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,105 @@
...
+            eventRepository.save(this)
```
marcus-bcl: Why do you need this?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,105 @@
...
+                date = LocalDate.now(),
+                startTime = ZonedDateTime.now(EuropeLondon),
+                staff = staff,
+                team = team,
+                provider = team.provider,
+                linkedContactId = contact.id,
+                notes = """
+                            ${DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm").format(LocalDateTime.now())}
+                            Enforcement Action: ${enforcementAction.description}
+                        """.trimIndent()
```
marcus-bcl: Why are we setting these notes on the enforcement action contact?  The story says:
> Update appointment notes (not enforcement action contact notes) to include enforcement action details

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,105 @@
...
+                "EnforcementActionCode",
+                enforcementActionCode
+            )
+        enforcementRepository.save(
+            Enforcement(
+                contact = contact,
+                action = enforcementAction,
+                responseDate = enforcementAction.responseByPeriod.let {
+                    contact.startTime?.plusDays(it ?: 0)
+                }
```
marcus-bcl: You're using `.let` without a null-check here - it might be clearer to remove it entirely?
```suggestion
                responseDate = contact.startTime?.plusDays(enforcementAction.responseByPeriod ?: 0)
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,105 @@
...
+                            ${DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm").format(LocalDateTime.now())}
+                            Enforcement Action: ${enforcementAction.description}
+                        """.trimIndent()
+            )
+        )
+        event?.run {
+            ftcCount = ftcCount?.plus(1)
+            eventRepository.save(this)
+            val ftcLimit = disposal?.type?.ftcLimit ?: return@run
+            if (ftcCount!! > ftcLimit && !contactRepository.enforcementReviewExists(
```
marcus-bcl: It looks like `ftcCount` has been made nullable again?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,105 @@
...
+                responseDate = enforcementAction.responseByPeriod.let {
+                    contact.startTime?.plusDays(it ?: 0)
+                }
+            )
+        )
+        contactRepository.save(
+            Contact(
+                person = person,
+                event = event,
+                requirement = requirement,
```
marcus-bcl: Similarly here, you don't need to look these up and store them as variables as you already have them on `contact`:
```suggestion
                person = contact.person,
                event = contact.event,
                requirement = contact.requirement,
```

projects/manage-supervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTestBase.kt
```diff
@@ -74,6 +75,13 @@ open class IntegrationTestBase {
     @Autowired
     lateinit var channelManager: HmppsChannelManager
 
+    @BeforeEach
+    fun drainChannel() {
+        val channel = channelManager.getChannel(topicName)
+        while (channel.receive() != null) { /* drain */
+        }
+    }
```
marcus-bcl: Do we need this?  We're not changing anything related to messaging are we?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
marcus-bcl: Also, are we copying all possible related entities here?  I think we'll also need to copy across the `licenceCondition` and `nsiId`.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
sambailey-moj: fixed

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
sambailey-moj: the 'bit at the top' is doing the validation - ie handling the nulls. Do we want to allow null staff and teams?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
sambailey-moj: fixed

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
sambailey-moj: fixed

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
sambailey-moj: fixed

projects/manage-supervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTestBase.kt
sambailey-moj: co-pilot put this logic here as I was getting a lot of intermittent test failures on repeated runs (not related to the change I'm making)... it seems to have fixed the problem I was experiencing - happy to remove it if you think it's not adding anything

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
marcus-bcl: I don't think it matters, we just need the enforcement action contact to match the passed in appointment contact (which we've already ensured is valid).

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -51,34 +44,36 @@ class ContactEnforcementService(
...
                             Enforcement Action: ${enforcementAction.description}
                         """.trimIndent()
-            )
-        )
-        event?.run {
-            ftcCount = ftcCount?.plus(1)
-            eventRepository.save(this)
+        })
+        contact.event?.run {
+            ftcCount = contactRepository.countFailureToComply(this).plus(1)
```
marcus-bcl: We shouldn't need the `plus(1)` here.  We've already saved the current appointment contact with `complied=false`, so it should be included in the count.

projects/manage-supervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/UpdateContactOutcomeIntegrationTest.kt
```diff
@@ -173,4 +175,87 @@ class UpdateContactOutcomeIntegrationTest : IntegrationTestBase() {
         }
         assertThat(ftcAfter, equalTo(ftcBefore + 1))
     }
+
+    @Test
+    fun `enforcement action creates linked contact with correct notes`() {
```
marcus-bcl: This test mentions creating the contact with correct notes, but doesn't actually check the notes?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Contact.kt
```diff
@@ -217,7 +218,7 @@ class Contact(
 
     @Lob
     var notes: String? = notes
-        private set
+        internal set
```
marcus-bcl: I think this was made private to prevent it being set directly - you should use the `appendNotes` function instead.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,103 @@
...
+                notes = null,
+                licenceCondition = contact.licenceCondition,
+                nsiId = contact.nsiId,
+            )
+        )
+        contactRepository.save(contact.apply {
+            this.notes = """
+                            ${DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm").format(LocalDateTime.now())}
+                            Enforcement Action: ${enforcementAction.description}
+                        """.trimIndent()
```
marcus-bcl: We need to append the notes here, otherwise we could unintentionally overwrite any previous notes that have been added to the contact.

projects/manage-supervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/IntegrationTestBase.kt
sambailey-moj: removed

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,97 @@
...
+                provider = contact.team?.provider,
+                linkedContactId = contact.id,
+                notes = null,
+                licenceCondition = contact.licenceCondition,
+                nsiId = contact.nsiId,
+            )
+        )
+        contactRepository.save(contact.apply {
+            this.attended = contactOutcome.outcomeAttendance
+            this.complied = contactOutcome.outcomeCompliantAcceptable
```
marcus-bcl: We should do this further up when we're actually setting the outcome - we need to do it regardless of whether we're setting an enforcement action.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -0,0 +1,97 @@
...
+    private val enforcementRepository: EnforcementRepository,
+    private val enforcementActionsRepository: EnforcementActionsRepository,
+) {
+    fun updateEnforcementActionForContact(contact: Contact, enforcementActionCode: String) {
+        val contactOutcome = contact.outcome.orNotFoundBy("contactId", contact.id)
+        val enforcementAction = enforcementActionsRepository.findByContactOutcomeId(contactOutcome.id)
+            .firstOrNull { it.code == enforcementActionCode }.orNotFoundBy(
+                "EnforcementActionCode",
+                enforcementActionCode
+            )
```
marcus-bcl: Minor thing - we should probably use `require` here to throw a 400 error instead of a 404:
```suggestion
        val enforcementAction = requireNotNull(enforcementActionsRepository.findByContactOutcomeId(contactOutcome.id)
            .firstOrNull { it.code == enforcementActionCode }) { "Enforcement action not valid for outcome" }
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
sambailey-moj: done

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
sambailey-moj: fixed

marcus-bcl (approved):

PR #6409, opened 2026-05-22
PI-4144 - Add validation to adjustment endpoints

joseph-bcl:


projects/community-payback-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/AdjustmentsIntegrationTest.kt
```diff
@@ -292,4 +292,148 @@ class AdjustmentsIntegrationTest @Autowired constructor(
...
+            )
+        }.andExpect { status { isBadRequest() } }.andExpect {
+            content {
+                string(
+                    org.hamcrest.Matchers.containsString(
+                        "Adjustment would result in negative remaining unpaid work minutes"
+                    )
+                )
+            }
+        }
```
marcus-bcl, line 326: Minor thing - you could use jsonPath to be more explicit here:
```suggestion
        }.andExpect { 
          status { isBadRequest() }
          content { jsonPath("$.message") { containsString("Adjustment would result in negative remaining unpaid work minutes") } }
        }
```

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/AdjustmentService.kt
marcus-bcl, line 17: I just noticed this isn't transactional 😱 

We should make this transactional, so that if any adjustment fails then the whole thing is rolled back.

That would then make life easier for your remaining time check, as you can just save the adjustment and validate the remainingMinutes result, rather than trying to project it in the Kotlin code.

PR #6412, opened 2026-05-26
PI-4149 Update custody key dates to stop adding the post sentence sup…

hazzaz-justice:


projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/Sentence.kt
```diff
@@ -74,8 +74,12 @@ data class DisposalType(
 
     @Column
     val requiredInformation: String,
+
+    @Column(name = "pss_rqmnt")
+    val pssRequirement: String? = null,
```
marcus-bcl: It might be simpler to use the converter here, to automatically map from Y/N to true/false
```suggestion
    @Column(name = "pss_rqmnt")
    @Convert(converter = YesNoConverter::class)
    val pssRequirement: Boolean? = null,
```

marcus-bcl (approved):

PR #6413, opened 2026-05-26
PI-4101 - new endpoint for non compliance detail MPoP 1

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ComplianceService.kt
```diff
@@ -114,6 +119,49 @@ class ComplianceService(
...
+    @Transactional
+    fun getPersonComplianceDetail(crn: String, months: Int): NonComplianceResponse {
+        require(months in 0..120) { "Months must be between 0 and 120" }
+        val summary = personRepository.getSummary(crn)
+        val events = eventRepository.findByPersonId(summary.id)
+        val currentSentences = events.filter { !it.isInactiveEvent() }
+        val allActiveSentenceContacts =
+            when (months) {
+                0 -> contactRepository.findByPersonIdAndEventIdIn(summary.id, currentSentences.map { it.id })
+                else -> contactRepository.findByPersonIdAndEventIdInAndCreatedDateTimeAfter(
```
marcus-bcl: Should we definitely be filtering on created date and not appointment date?  I think there is an index on contact_date, but not on created_datetime.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ComplianceService.kt
```diff
@@ -114,6 +119,49 @@ class ComplianceService(
...
+                    summary.id, currentSentences.map { it.id },
+                    ZonedDateTime.now().minusMonths(months.toLong())
+                )
+            }.filter { it.type.attendanceContact == true }
```
marcus-bcl: Would it be more efficient to do this check in the DB?  E.g. by adding `...AndTypeAttendanceContactTrue(` to the repository method.

marcus-bcl (dismissed):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ComplianceService.kt
sambailey-moj: good spot... we should be filtering on the contact date

marcus-bcl (approved):

PR #6423, opened 2026-05-27
PI-4154 Add last modified timestamp to personal details endpoint

hazzaz-justice:


projects/manage-my-community-sentence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/PersonalContact.kt
```diff
@@ -27,7 +28,8 @@ class PersonalContact(
     val emailAddress: String?,
     val startDate: LocalDate? = null,
     val endDate: LocalDate? = null,
-
+    @Column(name = "last_updated_datetime")
+    val lastUpdatedDatetime: LocalDateTime? = null,
```
marcus-bcl: Please could we use ZonedDateTime instead of LocalDateTime?  We made a decision a while back to use them wherever possible: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/doc/adr/0003-use-zoned-datetimes.md

Also, this column is not nullable so we can make it non-nullable here.

```suggestion
    val lastUpdatedDatetime: ZonedDateTime,
```

projects/manage-my-community-sentence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/address/PersonAddress.kt
```diff
@@ -32,7 +33,8 @@ class PersonAddress(
     @Convert(converter = YesNoConverter::class)
     val noFixedAbode: Boolean? = false,
     val endDate: LocalDate? = null,
-
+    @Column(name = "last_updated_datetime")
+    val lastUpdatedDatetime: LocalDateTime? = null,
```
marcus-bcl: Same here: 
```suggestion
    val lastUpdatedDatetime: ZonedDateTime,
```

projects/manage-my-community-sentence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Person.kt
```diff
@@ -29,6 +30,8 @@ class Person(
     val telephoneNumber: String? = null,
     @Column(name = "e_mail_address")
     val emailAddress: String? = null,
+    @Column(name = "last_updated_datetime")
+    val lastUpdatedDatetime: LocalDateTime? = null,
```
marcus-bcl: Same here: 
```suggestion
    val lastUpdatedDatetime: ZonedDateTime,
```

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #6427, opened 2026-05-28
PI-4124 - Add contact outcome, NS flag and last updated time to appointment endpoints

jamesmacbeth-unilink:


sambailey-moj (dismissed):

projects/manage-my-community-sentence-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/Appointment.kt
```diff
@@ -15,6 +16,9 @@ data class Appointment(
     val location: Address?,
     val attended: Boolean?,
     val complied: Boolean?,
+    val outcome: String?,
+    val nationalStandards: Boolean?,
+    val lastUpdatedAt: LocalDateTime?
```
marcus-bcl: Could we make this a ZonedDateTime?  Also, I don't think these two need to be nullable as they aren't nullable in the entities.
```suggestion
    val nationalStandards: Boolean,
    val lastUpdatedAt: ZocalDateTime,
```

marcus-bcl (approved):

PR #6435, opened 2026-06-02
PI-4178 populate enforcement flag on contact

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -140,6 +142,9 @@ class ContactLogService(
                 )
             }
 
+            if (savedContact.outcome != null) {
+                setEnforcementFlag(savedContact, savedContact.outcome)
+            }
```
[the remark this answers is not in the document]
sambailey-moj: sorted

libs/appointments/src/main/kotlin/uk/gov/justice/digital/hmpps/appointments/service/EnforcementService.kt
```diff
@@ -30,7 +30,11 @@ internal class EnforcementService(
 
     fun AppointmentContact.applyEnforcementAction(action: EnforcementAction) {
         enforcementActionId = action.id
-        enforcement = true
+        if (action.outstandingContactAction == true) {
+            enforcement = true
+        } else {
+            enforcement = null
+        }
```
marcus-bcl: We could simplify this to:

```suggestion
        enforcement = if (action.outstandingContactAction == true) true else null
```

libs/appointments/src/main/kotlin/uk/gov/justice/digital/hmpps/appointments/service/AppointmentService.kt
```diff
@@ -298,12 +299,22 @@ class AppointmentService internal constructor(
...
     ) = apply {
+        if (outcome == null && type.outcomeRequired == true) {
+            require(this.outcome == null) {
+                "Outcome is required for contact type ${type.code}"
+            }
+            enforcement = true
+            return@apply
+        }
+
         if (this.outcome?.code == outcome?.code) return@apply
```
marcus-bcl: This doesn't look quite right?  This line is to avoid making changes if the outcome is already set, so should come first:
```
if (this.outcome?.code == outcome?.code) return@apply
```

Also, we only require an outcome after the date/time has passed, so checking that flag here probably isn't right. We already validate that an outcome is always required for appointments in the `validateUpdates` method, which supersedes the `outcomeRequired` check.

In fact, I think we should move the initial setting of the enforcement up to the create method, as it's only based on type which can't be changed after creation (see the other comment).

```suggestion
        if (this.outcome?.code == outcome?.code) return@apply
```

libs/appointments/src/main/kotlin/uk/gov/justice/digital/hmpps/appointments/service/AppointmentService.kt
```diff
@@ -75,6 +75,7 @@ class AppointmentService internal constructor(
                     staff = staff[request.staffCode].orNotFoundBy("code", request.staffCode),
                     officeLocation = request.locationCode?.let { code -> locations[code].orNotFoundBy("code", code) },
                     type = types[request.typeCode].orNotFoundBy("code", request.typeCode),
+
```
marcus-bcl: We could just add something like this here:
```suggestion
enforcement = type.outcomeRequired == true
```
and add this at the top of the `map`:
```kotlin
val type = types[request.typeCode].orNotFoundBy("code", request.typeCode)
```

I think that makes it clearer that enforcement is initially based on the type, then later is based on the enforcement action as part of applying an outcome.

libs/appointments/src/main/kotlin/uk/gov/justice/digital/hmpps/appointments/service/AppointmentService.kt
```diff
@@ -75,7 +76,7 @@ class AppointmentService internal constructor(
                     staff = staff[request.staffCode].orNotFoundBy("code", request.staffCode),
                     officeLocation = request.locationCode?.let { code -> locations[code].orNotFoundBy("code", code) },
                     type = types[request.typeCode].orNotFoundBy("code", request.typeCode),
```
marcus-bcl: We could just make this `type = type,` now, to save us looking it up again.

sambailey-moj: > sorted

marcus-bcl (approved):

PR #6441, opened 2026-06-02
PI-4189 Update Contact GET endpoint to include alert field

hazzaz-justice:


sambailey-moj (dismissed):

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ScheduleService.kt
```diff
@@ -235,7 +235,8 @@ fun Contact.toActivity(noteId: Int? = null): Activity {
         component = requirement?.asComponent() ?: licenceCondition?.asComponent(),
         nsiId = nsiId,
         esupervisionId = eSupervisionId(),
-        externalReference = externalReference
+        externalReference = externalReference,
+        alert = (alert == true)
```
marcus-bcl: Is there any reason not to return null here, if the column is null?

My vague understanding is that:
* null - nobody has confirmed whether this is an alert or not
* true - someone has confirmed this is an alert
* false - someone has confirmed this is not an alert.

hazzaz-justice: Please check, it should be resolved now.

marcus-bcl (approved):

PR #6448, opened 2026-06-04
PI-4188 new endpoint for access control service

sambailey-moj:


libs/limited-access/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/LimitedAccess.kt
```diff
@@ -38,13 +46,20 @@ class Restriction(
...
     val end: LocalDateTime?,
 
     @Id
     @Column(name = "restriction_id")
     val id: Long
-)
+) {
+    // Secondary constructor preserving the pre-start signature used by existing call sites
+    constructor(person: LimitedAccessPerson, user: LimitedAccessUser, end: LocalDateTime?, id: Long) :
+        this(person, user, LocalDateTime.now(), end, id)
```
marcus-bcl: Would it be clearer to explicitly pass `start = LocalDate.now()` from the other call sites - rather than adding a second constructor and silently defaulting?

libs/limited-access/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/LimitedAccess.kt
```diff
@@ -38,13 +46,20 @@ class Restriction(
     @JoinColumn(name = "user_id")
     val user: LimitedAccessUser,
 
+    @Column(name = "restriction_time")
+    val start: LocalDateTime,
```
marcus-bcl: Same here

libs/limited-access/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/LimitedAccess.kt
```diff
@@ -93,6 +108,29 @@ interface UserAccessRepository : JpaRepository<LimitedAccessUser, Long> {
...
+        from Exclusion e
+        where e.person.crn = :crn
+        and (e.end is null or e.end > current_date)
+        """
+    )
+    fun getExclusionsForCrn(crn: String): List<ExclusionDetail>
+
+    @Query(
+        """
+        select r.user.username as username, r.start as since, r.end as until
```
marcus-bcl: Could we use consistent naming here?
```suggestion
        select e.user.username as username, e.start as since, e.end as until
        from Exclusion e
        where e.person.crn = :crn
        and (e.end is null or e.end > current_date)
        """
    )
    fun getExclusionsForCrn(crn: String): List<ExclusionDetail>

    @Query(
        """
        select r.user.username as username, r.start as since, r.end as until
```

libs/limited-access/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/LimitedAccess.kt
```diff
@@ -18,13 +20,19 @@ class Exclusion(
     @JoinColumn(name = "user_id")
     val user: LimitedAccessUser,
 
+    @Column(name = "exclusion_date")
+    val start: LocalDate,
```
marcus-bcl: Why do we map this one as a LocalDate and the other as a LocalDateTime?

We should probably update it to use `ZonedDateTime` everywhere for fields with a date and time component - even on JPA entities.  We made a call to do it a while back: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/main/doc/adr/0003-use-zoned-datetimes.md and we have a custom deserializer that ensures the correct timezone is used.

libs/limited-access/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/LimitedAccess.kt
```diff
@@ -4,7 +4,9 @@ import jakarta.persistence.*
 import org.hibernate.annotations.Immutable
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
+import java.time.LocalDate
```
marcus-bcl: Minor - unused import.  It might be worth enabling the IntelliJ "Optimise imports" precommit hook, or "Optimise imports on save" if you haven't already - https://github.com/ministryofjustice/hmpps-probation-integration-services#code-formatting

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #6452, opened 2026-06-05
PI-4193 - Add active recall NSI data to compliance/{crn} endpoint

joseph-bcl:
- Add active recall NSI data to compliance/{crn} endpoint


marcus-bcl (approved):
Do we need to handle the case where someone has multiple active Recall NSIs?

PR #6462, opened 2026-06-09
PI-4204 update probation status

sambailey-moj:


projects/person-search-index-from-delius/container/pipelines/person/statement.sql
```diff
@@ -275,6 +275,8 @@ SELECT json_object(
                                                        WHEN EXISTS(SELECT 1
                                                                    FROM EVENT e
                                                                    WHERE e.OFFENDER_ID = o.OFFENDER_ID
+                                                                     AND e.ACTIVE_FLAG = 1
+                                                                     AND e.SOFT_DELETED = 0
                                                                      AND NOT EXISTS(SELECT 1 FROM DISPOSAL d WHERE d.EVENT_ID = e.EVENT_ID))
```
marcus-bcl: Do we also need to check that the DISPOSAL is active and not soft-deleted?

marcus-bcl (approved):

PR #6463, opened 2026-06-09
PI-4207 checking for person exists

sambailey-moj:


projects/accredited-programmes-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseDetailService.kt
```diff
@@ -81,21 +81,33 @@ class CaseDetailService(
...
+                manager = Manager(
+                    staff = manager.staff.toProbationPractitioner { ldapTemplate.findEmailByUsername(it.username) },
+                    team = manager.team.toCodedValue(),
+                    probationDeliveryUnit = manager.team.localAdminUnit.probationDeliveryUnit.toCodedValue(),
+                    officeLocations = manager.team.officeLocations.map { it.toCodedValue() }
+                ),
+                probationDeliveryUnits = manager.team.pduOfficeLocations(),
+                eventNumber = manager.requirement.disposal.event.number
+            )
+        }
```
marcus-bcl: I think you can drop the `.let {` call here?
```suggestion
        return Requirement(
            manager = Manager(
                staff = manager.staff.toProbationPractitioner { ldapTemplate.findEmailByUsername(it.username) },
                team = manager.team.toCodedValue(),
                probationDeliveryUnit = manager.team.localAdminUnit.probationDeliveryUnit.toCodedValue(),
                officeLocations = manager.team.officeLocations.map { it.toCodedValue() }
            ),
            probationDeliveryUnits = manager.team.pduOfficeLocations(),
            eventNumber = manager.requirement.disposal.event.number
        )
```

projects/accredited-programmes-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CaseDetailService.kt
```diff
@@ -81,21 +81,33 @@ class CaseDetailService(
...
 
-    fun getLicenceCondition(id: Long) =
-        licenceConditionManagerRepository.findByLicenceConditionId(id)?.let { manager ->
+    fun getLicenceCondition(crn: String, id: Long): LicenceCondition {
+        val person = personRepository.getByCrn(crn)
+        val manager = licenceConditionManagerRepository.getLicenceConditionManagerByLicenceConditionId(id)
+        if (manager.licenceCondition.disposal.event.person.crn != person.crn) {
+            throw NotFoundException("LicenceCondition", "id", id)
+        }
+        return manager.let { manager ->
```
marcus-bcl: Same here, I don't think you need `.let` any more now that manager is not nullable.

marcus-bcl (approved):

PR #6469, opened 2026-06-10
PI-4217 - Cosso-and-delius bug fixes

joseph-bcl:
- Return multiple addresses using the office locations from a user's home area, setting the replyAddress from user preferences with a status of Default.
- Update Failure and Enforcement endpoint to use Contact.Description

### DIFF
--- a/projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/ResponsibleOfficerDetails.kt
+++ b/projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/ResponsibleOfficerDetails.kt
@@ -5,11 +5,12 @@ data class ResponsibleOfficerDetails(
     val emailAddress: String?,
     val telephoneNumber: String?,
     val probationArea: CodeAndDescription,
-    val replyAddress: OfficeAddress?,
+    val replyAddresses: List<OfficeAddress>?
 )
 
 data class OfficeAddress(
     val id: Long?,
+    val status: String?,
     val officeDescription: String?,
     val buildingName: String?,
     val buildingNumber: String?,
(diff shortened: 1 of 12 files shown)


projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/ResponsibleOfficerDetails.kt
```diff
@@ -5,11 +5,12 @@ data class ResponsibleOfficerDetails(
     val emailAddress: String?,
     val telephoneNumber: String?,
     val probationArea: CodeAndDescription,
-    val replyAddress: OfficeAddress?,
+    val replyAddresses: List<OfficeAddress>?
```
marcus-bcl: Should we default to an empty list here, rather than null?

projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ResponsibleOfficerService.kt
```diff
@@ -24,15 +23,18 @@ class ResponsibleOfficerService(
         val username = userRepository.findByStaffId(responsibleOfficer.staff.id)?.username
         val emailAddress = username?.let { ldapTemplate.findAttributeByUsername(username, "mail") }
         val telephoneNumber = username?.let { ldapTemplate.findAttributeByUsername(username, "telephoneNumber") }
-        val officeLocation = username
-            ?.let { ldapTemplate.findPreferenceByUsername(username, "replyAddress") }?.toLongOrNull()
-            ?.let { officeLocationRepository.findByIdOrNull(it) }
+        val homeArea = username?.let { ldapTemplate.findAttributeByUsername(username, "userHomeArea") }
+        val defaultReplyAddress =
+            username?.let { ldapTemplate.findPreferenceByUsername(it, "replyAddress")?.toLongOrNull() }
+        val officeLocations = homeArea?.let { officeLocationRepository.findAllByProbationAreaCode(it) }
```
marcus-bcl, line 29: In breach notice, they wanted us to return a 400 error when the user had no home area - do they want the same thing here?

marcus-bcl (dismissed):

projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ResponsibleOfficerService.kt
joseph-bcl: I've checked and we're happy to keep it as is (i.e. no 400 error). Suicide risk in the details service works this way.

marcus-bcl (approved):

PR #6472, opened 2026-06-11
PI-4200 Handle contacts not linked to event

hazzaz-justice:


projects/esupervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/messaging/Handler.kt
```diff
@@ -59,8 +59,12 @@ class Handler(
                 else -> throw IllegalArgumentException("Unexpected event type: ${notification.eventType}")
             }
         } catch (e: IgnorableMessageException) {
+            val eventName = when (notification.eventType) {
+                SETUP_REMOVED, SENTENCE_TERMINATED -> "CheckInSetupRemovedIgnored"
+                else -> "CheckInEventIgnored"
+            }
             telemetryService.trackEvent(
-                "CheckInEventIgnored",
+                eventName,
```
marcus-bcl: We should stick with the existing generic `CheckInEventIgnored` event name, we can rely on the `reason` property to distinguish it from other ignored check-in messages.

hazzaz-justice: Thanks, please recheck now

marcus-bcl (approved):

PR #6479, opened 2026-06-15
PI-4213 new sessions endpoint with multiple teams

sambailey-moj:


projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/ProvidersController.kt
```diff
@@ -57,6 +57,22 @@ class ProvidersController(
...
+        @RequestParam typeCode: List<String> = emptyList(),
+        @PageableDefault(page = 0, size = 10, sort = ["date", "projectName"]) pageable: Pageable
+    ) = providersService.getSessions(
+        teamCodes, startDate, endDate, typeCode, pageable.mapSorts(
+            "projectName" to """lower("projectName")""",
+            "date" to """"appointmentDate"""",
+            "allocatedCount" to """coalesce("allocatedCount", 0)""",
+            "outcomeCount" to """coalesce("outcomeCount", 0)"""
+        )
+    )
```
marcus-bcl: Could we move this out to a separate SessionsController and have the endpoint as `GET /sessions`, similar to how we have `GET /appointments`?  It feels like sessions are more of a first-class entity now in terms of Community Payback.

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProvidersService.kt
```diff
@@ -94,6 +94,38 @@ class ProvidersService(
...
+        endDate: LocalDate,
+        typeCodes: List<String>,
+        pageable: Pageable
+    ): SessionsResponse {
+        require(!endDate.isBefore(startDate)) { "endDate must be on or after startDate" }
+        require(ChronoUnit.DAYS.between(startDate, endDate) <= 7) { "Date range cannot be greater than 7 days" }
+
+        val teams = teamRepository.findTeamsByCodeIn(teamCodes)
+        val missingCodes = teamCodes - teams.map { it.code }.toSet()
+        if (missingCodes.isNotEmpty()) throw NotFoundException("Team", "code", missingCodes)
```
marcus-bcl: Could we use the existing `reportMissing` extension function for this?

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProvidersService.kt
```diff
@@ -94,6 +94,38 @@ class ProvidersService(
...
+        val sessions = unpaidWorkAppointmentRepository.getUnpaidWorkSessionDetails(
+            teamIds,
+            startDate,
+            endDate,
+            typeCodes,
+            pageable
+        )
+
+        return SessionsResponse(
+            sessions = deprecatedAllSessions.map { it.toModel() }.toList(),
```
marcus-bcl: Do they still need this deprecated list of un-pages sessions?

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/controller/ProvidersController.kt
sambailey-moj: done

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProvidersService.kt
sambailey-moj: That extension function only works on a Map, created new function that works on a list

marcus-bcl: I think you just need an associateBy - for the other services we've added an extension function to the repository e.g. https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/a586e311a01429e28000cff89fc6fb08ef2a34c2/projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/contact/ContactType.kt#L106-L107
```kotlin
    fun getByCodeIn(codes: List<String>) =
        codes.toSet().let { codes -> findByCodeIn(codes).associateBy { it.code }.reportMissing(codes) }
```

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProvidersService.kt
sambailey-moj: removed for this endpoint

projects/community-payback-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ProvidersService.kt
sambailey-moj: Actually that reportMissing doesnt throw a NotFoundException... isn't that what we require here?

marcus-bcl: It returns a 400 error with a list of the team codes that don't exist - I think that makes sense, as it's not that we haven't found a specific session, it's that they've provided invalid team codes.

sambailey-moj: cool.. sorted that then

marcus-bcl (approved):

PR #6491, opened 2026-06-16
PI-4216 Breach notice upload failure

hazzaz-justice:


projects/breach-notice-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/DocumentService.kt
marcus-bcl, line 1: I don't think we need the changes in this file - the test is helpful to add, but wrapping the delete call in `nullIfNotFound` already ignores the 404 error.

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #6536, opened 2026-06-23
PI-4196 - Final Third + EMED Date key date calculation

joseph-bcl:


projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/CustodyDateUpdateService.kt
```diff
@@ -66,25 +70,52 @@ class CustodyDateUpdateService(
...
+        envelope: OperativeSentenceEnvelope
+    ): LocalDate? =
+        envelope.sentenceEnvelopeLengthInDays?.let { days ->
+            sentenceDetail.conditionalReleaseDate?.let { crd ->
+                keyDateCalculator.calculatePresumptiveEMEndDate(crd, days, envelope.containsAnSDSPlusSentence)
+            }
+        }
+
+    private fun finalThirdDate(sentenceDetail: SentenceDetail, envelope: OperativeSentenceEnvelope): LocalDate? =
+        envelope.sentenceEnvelopeLengthInDays?.let { days ->
```
marcus-bcl: `sentenceEnvelopeLengthInDays` is non-nullable on their side, so we should be able to remove this check: https://github.com/ministryofjustice/calculate-release-dates-api/blob/ff5d35f96fbcf39f7db793e472d112488243a0c4/src/main/kotlin/uk/gov/justice/digital/hmpps/calculatereleasedatesapi/model/OperativeSentenceEnvelope.kt#L9

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/CustodyDateUpdateService.kt
```diff
@@ -66,25 +70,52 @@ class CustodyDateUpdateService(
...
+        )
 
     fun suspensionDateIfReset(sentenceDetail: SentenceDetail, custody: Custody): LocalDate? = custody.disposal
         ?.takeIf { it.type.determinateSentence }
         ?.let {
             val startDate = it.event.firstReleaseDate ?: sentenceDetail.conditionalReleaseDate ?: return null
             val endDate = sentenceDetail.sentenceExpiryDate ?: return null
-            return if (startDate < endDate) startDate.plusDays(DAYS.between(startDate, endDate) * 2 / 3) else null
+            keyDateCalculator.calculateSuspensionDateIfReset(startDate, endDate)
+        }
```
marcus-bcl: I wonder if it's worth moving all of this logic and null-checks into the KeyDateCalculator and removing the wrapper methods from CustodyDateUpdateService.kt, so it's all in one place?

So above we would end up with something like:
```kotlin
            custody.keyDate(SUSPENSION_DATE_IF_RESET.code, calculator.suspensionDate(sentenceDetail, custody)),
            custody.keyDate(
                PRESUMPTIVE_EM_END_DATE.code,
                calculator.presumptiveElectronicMonitoringEndDate(sentenceDetail, envelope)
            ),
            custody.keyDate(FINAL_THIRD_START_DATE.code, calculator.finalThirdDate(sentenceDetail, envelope)),
```

projects/custody-key-dates-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/custody/date/KeyDateCalculator.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * FTHRD Calculation = SLED - 1/3 sentence length
+     */
+    fun finalThirdDate(sentenceDetail: SentenceDetail, envelope: OperativeSentenceEnvelope): LocalDate? =
+        sentenceDetail.sentenceExpiryDate?.let { sed ->
+            val lengthInDays = envelope.sentenceEnvelopeLengthInDays
+            val deduction = (lengthInDays / 3.0).roundToLong()
+            sed.minusDays(deduction)
```
marcus-bcl: For suspension date, we use integer division to effectively always round down.

So for a 10-day sentence, the suspension date would be on day 6 (even though `10*2/3 = 6.66... ~ 7`).

However, this would set the final third date to be day 7 (`10 - (10/3.0) = 6.66... ~ 7`).

marcus-bcl: We might want to confirm with Andrew, but I think this should probably be consistent with the suspension date.

projects/custody-key-dates-and-delius/src/main/resources/application.yml
```diff
@@ -3,7 +3,7 @@ spring:
   jackson:
     default-property-inclusion: non_null
   jpa:
-    #    hibernate.ddl-auto: validate # TODO re-enable this after "sds-plus-flag-enabled" feature flag is enabled
+    hibernate.ddl-auto: validate
```
marcus-bcl: I think you still need this to be disabled, so that you can deploy to prod ahead of the Delius DB deployment.

PR #6553, opened 2026-06-24
PI-4288 tweak logic to deliusManaged field

sambailey-moj:
MPoP will be managing non-compliant outcomes moving forward so little tweak to the 'deliusManaged' flag in the schedule

### DIFF
--- a/projects/manage-supervision-and-delius/src/test/kotlin/uk/gov/justice/digital/hmpps/service/ScheduleServiceTest.kt
+++ b/projects/manage-supervision-and-delius/src/test/kotlin/uk/gov/justice/digital/hmpps/service/ScheduleServiceTest.kt
@@ -2,7 +2,6 @@ package uk.gov.justice.digital.hmpps.service
 
 import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers.equalTo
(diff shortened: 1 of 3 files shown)


projects/manage-supervision-and-delius/src/test/kotlin/uk/gov/justice/digital/hmpps/service/ScheduleServiceTest.kt
```diff
@@ -110,4 +110,22 @@ internal class ScheduleServiceTest {
...
+    fun `deliusManaged is true when contact type is unknown (not in CreateAppointment types)`() {
+        val contact = ContactGenerator.FIRST_NON_APPT_CONTACT
+        assertThat(contact.toActivity().deliusManaged, equalTo(true))
+    }
+
+    @Test
+    fun `deliusManaged is true when requirement mainCategory code is F`() {
+        val contact = ContactGenerator.FIRST_NON_APPT_CONTACT
+        assertThat(contact.toActivity().deliusManaged, equalTo(true))
+    }
```
marcus-bcl, line 143: Are these tests doing the same thing?  Are we actually checking a case where the requirement mainCategory is F?

marcus-bcl (approved):

PR #6554, opened 2026-06-24
PI-4130 new endpoint for multiple enforcement actions

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -306,4 +310,51 @@ class ContactLogService(
...
+        latestAction?.let { contact.latestEnforcementAction = it }
+    }
+
+    private fun createEnforcementContact(contact: Contact) {
+        val enforcementContact = Contact(
+            person = contact.person,
+            notes = "",
+            linkedContactId = contact.id,
+            type = contactTypeRepository.getContactType("AROM"),
+            date = LocalDate.now(),
```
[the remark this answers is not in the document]
sambailey-moj: intended logic

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -306,4 +310,53 @@ class ContactLogService(
             setEnforcementFlag(contact)
         }
     }
+
+    @Transactional
+    fun updateEnforcementContactOutcome(contactId: Long, request: UpdateEnforcementActions) {
+        val contact = contactRepository.getContact(contactId)
+        require(contact.outcome != null) { "Contact requires outcome" }
```
marcus-bcl: Do we need to check that `contact.outcome?.complied == false` ?  I don't think you can add enforcement actions to contacts with compliant outcomes.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -306,4 +310,53 @@ class ContactLogService(
...
+        contact.enforcementFlag = true
+        var latestAction: EnforcementAction? = null
+        val outcomeId = contact.outcome!!.id
+        val validActionIds = enforcementActionsRepository.findByContactOutcomeId(outcomeId).map { it.id }.toSet()
```
marcus-bcl: Would it be better to limit this to just the action codes they're requesting, rather than fetching the whole lot?
```suggestion
        val validActions = enforcementActionsRepository.findByContactOutcomeIdAndCodeIn(outcomeId, request.enforcementActions.map { it.code }.toSet())
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -306,4 +310,53 @@ class ContactLogService(
...
+        request.enforcementActions.forEach { ea ->
+            val enforcementAction = enforcementActionsRepository.getEnforcementActionByCode(ea.code)
+            require(validActionIds.contains(enforcementAction.id))
+            { "Enforcement action must be valid for outcome" }
```
marcus-bcl: With the change in the above comment you should be able to avoid the N+1 query here by removing the repository call:
```suggestion
            val enforcementAction = validActions.singleOrNull { it.code == ea.code } 
                ?: error("Enforcement action must be valid for outcome")
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -306,4 +310,53 @@ class ContactLogService(
...
+            createEnforcementContact(contact)
+            enforcementRepository.save(enforcement)
+            contact.appendNotes(
+                """
+                ${DateTimeFormatter.ofPattern("dd/MM/yyyy HH:mm").format(LocalDateTime.now())}
+                Enforcement Action: ${enforcementAction.description}
+                """.trimIndent()
+            )
+        }
+        latestAction?.let { contact.latestEnforcementAction = it }
```
marcus-bcl: Minor thing - you could avoid the mutable var by just setting the latest action explicitly to the last in the list.
```suggestion
        contact.latestEnforcementAction = validActions.single { it.code == request.enforcementActions.last().code }
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -306,4 +310,53 @@ class ContactLogService(
...
+        val validActionIds = enforcementActionsRepository.findByContactOutcomeId(outcomeId).map { it.id }.toSet()
+        request.enforcementActions.forEach { ea ->
+            val enforcementAction = enforcementActionsRepository.getEnforcementActionByCode(ea.code)
+            require(validActionIds.contains(enforcementAction.id))
+            { "Enforcement action must be valid for outcome" }
+            latestAction = enforcementAction
+            val enforcement = Enforcement(
+                contact = contact,
+                action = enforcementAction,
+                responseDate = contact.startTime?.plusDays(enforcementAction.responseByPeriod ?: 0)
```
marcus-bcl: Should `responseDate` be set to null if `responseByPeriod` is null?  This sets it to the current date.  What does Delius do?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -306,4 +310,53 @@ class ContactLogService(
...
+        }
+        latestAction?.let { contact.latestEnforcementAction = it }
+    }
+
+    private fun createEnforcementContact(contact: Contact) {
+        val enforcementContact = Contact(
+            person = contact.person,
+            notes = "",
+            linkedContactId = contact.id,
+            type = contactTypeRepository.getContactType("AROM"),
```
marcus-bcl: This is hard-coded to the AROM contact type, but it should use the contact type linked to the enforcement action code that was passed in - using `r_enforcement_action.contact_type_id`

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -314,22 +314,24 @@ class ContactLogService(
...
         val contact = contactRepository.getContact(contactId)
-        require(contact.outcome != null) { "Contact requires outcome" }
+        require(contact.outcome?.outcomeCompliantAcceptable == false) { "Contact requires outcome to be non compliant" }
         contact.enforcementFlag = true
-        var latestAction: EnforcementAction? = null
         val outcomeId = contact.outcome!!.id
-        val validActionIds = enforcementActionsRepository.findByContactOutcomeId(outcomeId).map { it.id }.toSet()
+        val validActions = enforcementActionsRepository.findByContactOutcomeIdAndCodeIn(
+            outcomeId,
+            request.enforcementActions.map { it.code }.toList()
```
marcus-bcl: Would it be better to pass a Set here, to avoid passing the same code multiple times?  As it stands, if they pass duplicate codes, then we will fail on the `singleOrNull` check later on.

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -314,22 +314,24 @@ class ContactLogService(
...
+        )
         request.enforcementActions.forEach { ea ->
-            val enforcementAction = enforcementActionsRepository.getEnforcementActionByCode(ea.code)
-            require(validActionIds.contains(enforcementAction.id))
-            { "Enforcement action must be valid for outcome" }
-            latestAction = enforcementAction
+            val enforcementAction = validActions.singleOrNull { it.code == ea.code }
+                ?: error("Enforcement action must be valid for outcome")
+            val responseDate =
+                enforcementAction.responseByPeriod.let { if (it == null) null else contact.startTime?.plusDays(it) }
```
marcus-bcl: Any reason not to use the [safe call operator](https://kotlinlang.org/docs/null-safety.html#safe-call-operator) (`?.`) here?
```suggestion
            val responseDate = enforcementAction.responseByPeriod
                 ?.let { contact.startTime?.plusDays(it) }
```

marcus-bcl (approved):

PR #6560, opened 2026-06-25
PI-4191 DTS: WRA - Basic Details Page Integrations

hazzaz-justice:


projects/warrant-risk-assessment-and-delius/src/dev/kotlin/uk/gov/justice/digital/hmpps/data/DataLoader.kt
```diff
@@ -10,6 +10,21 @@ class DataLoader(dataManager: DataManager) : BaseDataLoader(dataManager) {
     override fun systemUser() = UserGenerator.AUDIT_USER
 
     override fun setupData() {
-        // Perform dev/test database setup here, using the `save()` and `saveAll()` methods
+        save(UserGenerator.AUDIT_USER)
+        save(PersonGenerator.TITLE)
```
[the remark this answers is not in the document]
hazzaz-justice: Fixed

projects/warrant-risk-assessment-and-delius/deploy/values-dev.yml
```diff
@@ -1,4 +1,4 @@
-enabled: false # TODO set this to true when you're ready to deploy your service
+enabled: true
```
marcus-bcl: You can just remove this whole line - it defaults to true.  Same for the values-preprod.yml file too.

```suggestion
```

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/config/AsyncApiConfig.kt
marcus-bcl, line 1: This file is still required for when we add the async message consumer

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Address.kt
```diff
@@ -0,0 +1,57 @@
...
+    @Column(name = "end_date")
+    val endDate: LocalDate? = null,
+
+    @Column(columnDefinition = "number")
+    @Convert(converter = NumericBooleanConverter::class)
+    val softDeleted: Boolean = false,
+)
+
+interface AddressRepository : JpaRepository<Address, Long> {
+    fun findByPersonIdAndEndDateIsNull(personId: Long): List<Address>
```
marcus-bcl: What if the end date is in the future?  We've handled this with a SQL restriction in other services: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/467366d446585a407f2769aca5e4b02cf9ff57a1/projects/pre-sentence-reports-to-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/PersonAddress.kt#L15

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/BasicDetails.kt
```diff
@@ -0,0 +1,56 @@
...
+    val employerName: Name,
+    val employerAddress: EmployerAddress?,
+    val telephoneNumber: String?,
+    val mobileNumber: String?,
+)
+
+data class EmployerAddress(
+    val id: Long,
+    val status: String?,
+    @JsonInclude(JsonInclude.Include.ALWAYS)
```
marcus-bcl: Why do you need this annotation?

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/BasicDetailsService.kt
```diff
@@ -0,0 +1,73 @@
...
+class BasicDetailsService(
+    private val personRepository: PersonRepository,
+    private val addressRepository: AddressRepository,
+    private val personalContactRepository: PersonalContactRepository,
+    private val contactRepository: ContactRepository,
+) {
+    fun getBasicDetails(crn: String): BasicDetails {
+        val person = personRepository.getPerson(crn)
+        val addresses = addressRepository.findByPersonIdAndEndDateIsNull(person.id)
+            .sortedWith(compareBy<Address>({ it.status.code }, { it.id }))
```
marcus-bcl: Do we need to sort these by status and ID?  If we're sorting we should do it in the database query, but I'm not sure if we need it at all here?

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/App.kt
```diff
@@ -1,10 +1,8 @@
 package uk.gov.justice.digital.hmpps
 
-import com.asyncapi.kotlinasyncapi.springweb.EnableAsyncApi
 import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.runApplication
 
-@EnableAsyncApi
```
marcus-bcl, line 7: We shouldn't remove this - we still want to generate the AsyncApi docs.

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/PersonalContact.kt
```diff
@@ -0,0 +1,100 @@
...
+    val softDeleted: Boolean = false,
+)
+
+interface PersonalContactRepository : JpaRepository<PersonalContact, Long> {
+    @Query(
+        """
+        select pc from PersonalContact pc
+        where pc.personId = :personId
+        and pc.relationshipType.code = 'CE'
+        and pc.endDate is null
```
marcus-bcl, line 96: Do we need to account for end date in the future?
```suggestion
        and (pc.endDate is null or pc.endDate > current_date)
```
Maybe it would be better to put that in a SQLRestriction on the entity to keep the query clean?

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/BasicDetails.kt
hazzaz-justice: https://dsdmoj.atlassian.net/browse/PI-4191 shows "buildingName": null field within address

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/BasicDetailsService.kt
hazzaz-justice: we dont need to, basically copilot suggested to keep API output deterministic as the api exposed externally and also for the integration test. if we dont need to, I can just revert back and relax the test conditions, please let me know.

hazzaz-justice (commented):
Please recheck when you are available next @marcus-bcl

projects/warrant-risk-assessment-and-delius/deploy/database/access.yml
```diff
@@ -2,6 +2,15 @@ database:
...
     password_key: /warrant-risk-assessment-and-delius/db-password
+    tables:
+      - audited_interaction
+      - offender
+      - offender_address
+      - r_standard_reference_list
+      - personal_contact
+      - address
+      - contact
+      - r_contact_type
```
marcus-bcl: This list is for tables we need to write to, but I don't think we need to write to any of these?

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/BasicDetails.kt
marcus-bcl: Thanks, we don't need that - we can just omit the null field to stay consistent with the rest of the app.

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/entity/Address.kt
marcus-bcl: You're still checking `AndEndDateIsNull` here, do you still need to now that you've added the SQL restriction?

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/BasicDetailsService.kt
marcus-bcl: If we're going to do any sorting, we should do it in the database anyway.  Maybe it would be better to update the repository method to something like `findByPersonIdOrderByStartDate` ?

settings.gradle.kts
```diff
@@ -63,6 +63,7 @@ include(
     "tier-to-delius",
     "unpaid-work-and-delius",
     "workforce-allocations-to-delius",
+    "warrant-risk-assessment-and-delius",
```
marcus-bcl: Should this entry be included twice in the list?

projects/warrant-risk-assessment-and-delius/deploy/values-preprod.yml
```diff
@@ -1,4 +1,4 @@
-enabled: false # TODO set this to true when you're ready to deploy your service
+enabled: true
```
marcus-bcl: Same here, you can remove this
```suggestion
```

marcus-bcl (approved):

PR #6561, opened 2026-06-25
PI-4222 adding in current tier, enable search by tier code via search…

sambailey-moj:
… endpoint

### DIFF
--- a/projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
+++ b/projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
@@ -134,7 +134,7 @@ class UserService(
             caseloadRepository.findSentenceTypesForStaff(staff.id).map { KeyPair(it.code.trim(), it.description) }
         val contactTypes =
             caseloadRepository.findContactTypesForStaff(staff.id).map { KeyPair(it.code.trim(), it.description) }
-        val tierTypes = referenceDataRepository.findByDatasetCode("TIER").map { KeyPair(it.code, it.description) }
+        val tierTypes = referenceDataRepository.findByDatasetCode("TIER").map { KeyPair(it.code.trim(), it.description) }
         val userAccess = userAccessService.userAccessFor(username, caseload.content.map { it.crn })
 
         return StaffCaseload(
@@ -356,9 +356,6 @@ class UserService(
             .map { Staff(name = Name(forename = it.forename, surname = it.surname), code = it.code) }
         return TeamStaff(provider = provider, staff = staff)
     }
-
-    fun getTiers() =
-        referenceDataRepository.findByDatasetCode("TIER").map { CodeAndDescription(it.code, it.description) }
 }
 
 fun CaseloadItem.toStaffCase(caseAccess: CaseAccess? = null) = if (caseAccess.isLao()) {
--- a/projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
+++ b/projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
@@ -357,7 +357,8 @@ class UserService(
         return TeamStaff(provider = provider, staff = staff)
     }
 
-    fun getTiers() = referenceDataRepository.findByDatasetCode("TIER").map { CodeAndDescription(it.code, it.description) }
+    fun getTiers() =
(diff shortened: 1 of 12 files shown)


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -351,6 +356,9 @@ class UserService(
             .map { Staff(name = Name(forename = it.forename, surname = it.surname), code = it.code) }
         return TeamStaff(provider = provider, staff = staff)
     }
+
+    fun getTiers() =
+        referenceDataRepository.findByDatasetCode("TIER").map { CodeAndDescription(it.code, it.description) }
```
marcus-bcl: Is this used?

projects/manage-supervision-and-delius/src/test/kotlin/uk/gov/justice/digital/hmpps/service/UserServiceTest.kt
```diff
@@ -101,6 +100,9 @@ internal class UserServiceTest {
     @Mock
     lateinit var providerRepository: ProviderRepository
 
+    @Mock
+    lateinit var referenceDataRepository: uk.gov.justice.digital.hmpps.integrations.delius.referencedata.entity.ReferenceDataRepository
```
marcus-bcl: Any reason not to import this?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -124,12 +127,14 @@ class UserService(
...
             nextContactCode = searchFilter.nextContactCode?.trim()?.uppercase(),
-            sentenceCode = searchFilter.sentenceCode?.trim()?.uppercase()
+            sentenceCode = searchFilter.sentenceCode?.trim()?.uppercase(),
+            tierCode = searchFilter.tierCode?.trim()?.uppercase()
         )
         val sentenceTypes =
             caseloadRepository.findSentenceTypesForStaff(staff.id).map { KeyPair(it.code.trim(), it.description) }
         val contactTypes =
             caseloadRepository.findContactTypesForStaff(staff.id).map { KeyPair(it.code.trim(), it.description) }
+        val tierTypes = referenceDataRepository.findByDatasetCode("TIER").map { KeyPair(it.code, it.description) }
```
marcus-bcl: This returns all possible tier values, including V2 and V3 and legacy values - should it not return the tier values that are available within the current caseload?  Do they even need the full list of tiers, as there are only a few and they're well known across services?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
```diff
@@ -124,12 +127,14 @@ class UserService(
             staffId = staff.id,
             nameOrCrn = searchFilter.nameOrCrn?.trim()?.lowercase(),
             nextContactCode = searchFilter.nextContactCode?.trim()?.uppercase(),
-            sentenceCode = searchFilter.sentenceCode?.trim()?.uppercase()
+            sentenceCode = searchFilter.sentenceCode?.trim()?.uppercase(),
+            tierCode = searchFilter.tierCode?.trim()?.uppercase()
```
marcus-bcl, line 130: Do we need to map this from the canonical tier code (e.g. A-G, A1, B2, ...) to the Delius tier code?  Would it be better to maintain an enum of v2 and v3 tiers with mappings to the Delius codes?

marcus-bcl (commented):
It's a shame we can't point them at the standard Tier API for this, but understand if they need to filter the caseload using the tier.  Unfortunately it means they won't be able to show V3 tiers in the caseload for a while, as Delius will continue to store the V2 tier in `offender.current_tier` until September (https://dsdmoj.atlassian.net/browse/PI-4198)

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
sambailey-moj: I think this is the right approach, to make it behave like the other filters.. ie. as you suggest return the values for the current caseload

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/UserService.kt
sambailey-moj: should have been deleted now

PR #6564, opened 2026-06-26
PI-4299 single enforcement for a contact

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/integrations/delius/overview/entity/Contact.kt
```diff
@@ -32,6 +32,7 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.ZonedDateTime
 import java.time.format.DateTimeFormatter
+import java.util.Optional
```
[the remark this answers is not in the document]
sambailey-moj: fixed

projects/manage-supervision-and-delius/src/integrationTest/kotlin/uk/gov/justice/digital/hmpps/UpdateEnforcementActionsIntegrationTest.kt
```diff
@@ -100,7 +100,7 @@ class UpdateEnforcementActionsIntegrationTest : IntegrationTestBase() {
         val linkedContactCountAfter = contactRepository.findByLinkedContactIdOrderByDateDesc(contactId)
             .count { it.type.code == expectedContactTypeCode }
 
-        assertThat(enforcementCountAfter, equalTo(enforcementCountBefore + 2))
+        assertThat(enforcementCountAfter, equalTo(enforcementCountBefore + 1))
```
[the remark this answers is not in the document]
sambailey-moj: fixed

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -271,17 +271,17 @@ class ContactLogService(
...
+                    "enforcements" to contact.enforcement.let {
                         mapOf(
-                            "id" to it.id,
-                            "action" to it.action?.code,
-                            "responseDate" to it.responseDate
+                            "id" to it?.id,
+                            "action" to it?.action?.code,
+                            "responseDate" to it?.responseDate
                         )
                     }.let { objectMapper.writeValueAsString(it) }
```
marcus-bcl: Use `?.let` to handle the null case up-front:

```suggestion
                    "enforcements" to contact.enforcement?.let {
mapOf(
                            "id" to it.id,
                            "action" to it.action.code,
                            "responseDate" to it..responseDate
                       )
                    }?.let { objectMapper.writeValueAsString(it) }
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -271,17 +271,17 @@ class ContactLogService(
...
                 )
             )
-            enforcementRepository.deleteAll(contact.enforcements)
-            contact.enforcements = mutableListOf()
+            contact.enforcement?.let { enforcementRepository.delete(it) }
+            contact.enforcement = null
```
marcus-bcl, line 284: Are we supposed to delete the enforcement here?  What does the Delius code do when updating a contact with enforcement action from non-compliant to compliant?

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactEnforcementService.kt
```diff
@@ -23,13 +23,20 @@ class ContactEnforcementService(
...
+                Enforcement(
+                    contact = contact,
+                    action = enforcementAction,
+                    responseDate = responseDate
+                )
             )
-        )
+        } else {
+            contact.enforcement!!.action = enforcementAction
+            contact.enforcement!!.responseDate = contact.startTime?.plusDays(enforcementAction.responseByPeriod ?: 0)
```
marcus-bcl: This is adding 0 days if there is no response period, but it should set to null same as the above branch

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
sambailey-moj: Delius deletes the enforcement
Created a contact, marked non-compliant, then marked it as compliant.... Delius zapped the enforcement

marcus-bcl (dismissed):

marcus-bcl (approved):

PR #6565, opened 2026-06-26
Pi 4205

hazzaz-justice:


projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/model/ResponsibleOfficerDetails.kt
```diff
@@ -0,0 +1,29 @@
...
+data class CodeAndDescription(
+    val code: String,
+    val description: String,
+)
+
+data class OfficeAddress(
+    val id: Long?,
+    val status: String?,
+    val officeDescription: String?,
+    @JsonInclude(JsonInclude.Include.ALWAYS)
```
marcus-bcl: Is this the same as the last PR?  Can we remove this if we don't need it.

projects/warrant-risk-assessment-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ResponsibleOfficerService.kt
```diff
@@ -0,0 +1,70 @@
...
+                surname = staff.surname,
+            ),
+            telephoneNumber = telephoneNumber,
+            probationArea = CodeAndDescription(
+                code = probationArea.code,
+                description = probationArea.description,
+            ),
+            replyAddress = replyAddress,
+        )
+    }
```
marcus-bcl, line 39: You might be able to simplify this down using extension functions.  It seems to have the exact same logic as the existing endpoint in the cosso-and-delius project: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/eb2c21a1bed0deeeaaba791956a6843547317e47/projects/cosso-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ResponsibleOfficerService.kt#L21

Would it be better to copy the logic from there, so that we know it's implemented consistently?

projects/warrant-risk-assessment-and-delius/build.gradle.kts
```diff
@@ -18,6 +18,9 @@ dependencies {
     implementation("org.jetbrains.kotlin:kotlin-stdlib-jdk8")
     implementation("tools.jackson.module:jackson-module-kotlin")
     implementation(libs.springdoc)
+    implementation("org.springframework.boot:spring-boot-starter-data-ldap")
+
+    dev("com.unboundid:unboundid-ldapsdk")
```
marcus-bcl, line 23: When adding the LDAP library, we also need to remember to update the Helm deploy files with the LDAP connection details - for example: https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/c1515660d739530b7af3038ea55debac22dd4469/projects/cosso-and-delius/deploy/values.yaml#L16-L18

projects/warrant-risk-assessment-and-delius/deploy/values-preprod.yml
```diff
@@ -1,6 +1,6 @@
 generic-service:
   ingress:
-    host: warrant-risk-assessment-and-delius-preprod.hmpps.service.justice.gov.uk
+    host: warrant-risk-assessment-api-preprod.hmpps.service.justice.gov.uk
```
marcus-bcl: Do these changes to values-preprod and values-prod need to be reverted?

projects/warrant-risk-assessment-and-delius/src/main/resources/application.yml
```diff
@@ -24,22 +24,27 @@ spring:
...
         token-uri: http://localhost:${dev.wiremock.port}/auth/oauth/token
+  ldap:
+    base: ou=Users,dc=moj,dc=com
+    base-environment:
+      java.naming.ldap.derefAliases: never
   threads.virtual.enabled: true
 
 oauth2.roles:
+  - PROBATION_API__WARRANT_RISK_ASSESSMENT__RESPONSIBLE_OFFICER
   - PROBATION_API__WARRANT_RISK_ASSESSMENT__BASIC_DETAILS
```
marcus-bcl: Could we merge these role names into a single `PROBATION_API__WARRANT_RISK_ASSESSMENT__CASE_DETAILS` ?  Just so that the client team doesn't need to request a new role for each endpoint.

projects/warrant-risk-assessment-and-delius/build.gradle.kts
```diff
@@ -35,4 +38,8 @@ configure<ClassPathExtension> {
         "**/entity/**",
         "**/AppKt.class"
     )
+    sonarExclusions = listOf(
+        "**/entity/ResponsibleOfficer.kt",
+        "**/service/ResponsibleOfficerService.kt"
+    )
```
marcus-bcl: We don't need these exclusions, we can still merge even when Sonar tells us there is some duplication - it is expected in this repo as we have some overlap between projects.
```suggestion
```

PR #6569, opened 2026-06-30
PI-4300 update contacts with no outcome

sambailey-moj:


projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -261,17 +261,25 @@ class ContactLogService(
...
+            contactTypeRepository.findSelectableOutcomesByTypeCode(contactType)
+                .firstOrNull { it.code == code }
+                .orNotFoundBy("code", code)
+        }
+        if (contactOutcome == null) {
+            require(contact.outcome == null) { "outcomeCode cannot be null when the contact already has an outcome" }
+        }
+        if (request.enforcementActionCode != null && contactOutcome == null) {
+            throw InvalidRequestException("outcome is required when an enforcement action is provided")
+        }
```
marcus-bcl: Should we use `require` consistently, if we're using it above too?
```suggestion
        require (request.enforcementActionCode == null || contactOutcome != null) { "Outcome is required when an enforcement action is provided" }
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -261,17 +261,25 @@ class ContactLogService(
...
-            .firstOrNull { it.code == request.outcomeCode }
-            .orNotFoundBy("code", request.outcomeCode)
+        val contactOutcome = request.outcomeCode?.let { code ->
+            contactTypeRepository.findSelectableOutcomesByTypeCode(contactType)
+                .firstOrNull { it.code == code }
+                .orNotFoundBy("code", code)
+        }
+        if (contactOutcome == null) {
+            require(contact.outcome == null) { "outcomeCode cannot be null when the contact already has an outcome" }
+        }
```
marcus-bcl, line 271: Could this be simplified into a single check? E.g.
```suggestion
        require(contactOutcome != null || contact.outcome == null) { "Contact outcome cannot be removed" }
```

projects/manage-supervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/ContactLogService.kt
```diff
@@ -297,17 +306,16 @@ class ContactLogService(
...
-                contactEnforcementService.updateEnforcementActionForContact(contact, request.enforcementActionCode)
-            setEnforcementFlag(contact, appliedAction)
-        } else {
-            setEnforcementFlag(contact)
-        }
+        val appliedAction = if (contactOutcome != null) {
+            request.enforcementActionCode?.let { code ->
+                contactEnforcementService.updateEnforcementActionForContact(contact, code)
+            }
+        } else null
```
marcus-bcl: Would it be simpler to combine these null-checks?
```suggestion
        val appliedAction = if (contactOutcome != null && request.enforcementActionCode != null) {
                contactEnforcementService.updateEnforcementActionForContact(contact, request.enforcementActionCode)
        } else null
```

marcus-bcl (approved):

PR #6571, opened 2026-06-30
PI-4197 removal outcome codes

sambailey-moj:


projects/esupervision-and-delius/src/main/kotlin/uk/gov/justice/digital/hmpps/service/CheckInService.kt
```diff
@@ -71,7 +71,8 @@ class CheckInService(
         }
         if (contact.event == null) throw IgnorableMessageException("Event not found for setup removal")
         require(domainEvent matches contact) { "Case details mismatch" }
-        contact.outcome = contactOutcomeRepository.getByCode(SETUP_REMOVED)
+        val outcomeCode = domainEvent.additionalInformation["outcomeCode"]?.toString()?.takeIf { it.isNotBlank() } ?: SETUP_REMOVED
+        contact.outcome = contactOutcomeRepository.getByCode(outcomeCode)
```
marcus-bcl: We don't need to handle this being null, as they'll be sending it in prod by the time we deploy our changes.  

Also, we should add an extension function to Handler.kt to be consistent with the other items we're getting from additionalInformation.

```suggestion
        contact.outcome = contactOutcomeRepository.getByCode(domainEvent.outcomeCode)
```

---

Probably worth validating that the outcome code is one of the expected values too:

* Handler.kt
```kotlin
val HmppsDomainEvent.outcomeCode get() = requireNotNull(additionalInformation["outcomeCode"]).toString().also {
    require(it in listOf(SETUP_REMOVED, MANUAL_STOP, NO_ACTIVE_EVENTS, IN_RESET)) { "Unexpected outcome code: $it" }
}

```

marcus-bcl (approved):

